PR #24, opened 2023-07-06
feat: allow for dynamic init with `SETUP_SCRIPT`

kylecarbs:


bpmct (approved):
Nice! This LGTM. Should we do a similar thing with envbox? [This](https://coder.com/docs/v2/latest/templates/docker-in-workspaces#systemd-in-docker) is how we're documenting how to run systemd as PID 1 with envbox which isn't great.

If so, I'll make an issue

kylecarbs:
@bpmct I'm not sure if envbox requires any background job processing - if so then we can't do it this way, but if not we totally can.

PR #30, opened 2023-07-13
fix: Use fallback image if devcontainer.json doesn't specify an image or Dockerfile

aaronlehmann:
We might want a simple devcontainer.json that does something like reference a feature or set environment variables, but leaves the image as the default (fallback) image.

### DIFF
--- a/devcontainer/devcontainer.go
+++ b/devcontainer/devcontainer.go
@@ -56,11 +56,22 @@ type Compiled struct {
 	Env  []string
 }
 
+// HasImage returns true if the devcontainer.json specifies an image.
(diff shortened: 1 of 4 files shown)


aaronlehmann:
cc @kylecarbs

kylecarbs (approved):
@aaronlehmann it'd be awesome to have an integration test for this.

I'm happy to add one too - the only dependency is Docker!

aaronlehmann:
> I'm happy to add one too - the only dependency is Docker!

That would be great!

kylecarbs:
@aaronlehmann added!

PR #33, opened 2023-07-18
feat: allow for configurable ignore paths

kylecarbs:
This resolves coder/coder#8540.

### DIFF
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -137,6 +137,10 @@ type Options struct {
 	// and pulling from container registries.
 	Insecure bool `env:"INSECURE"`
 
+	// IgnorePaths is a comma separated list of paths
+	// to ignore when building the workspace.
+	IgnorePaths []string `env:"IGNORE_PATHS"`
+
 	// SkipRebuild skips building if the MagicFile exists.
 	// This is used to skip building when a container is
 	// restarting. e.g. docker stop -> docker start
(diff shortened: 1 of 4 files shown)


envbuilder.go
```diff
@@ -464,17 +471,24 @@ func Run(ctx context.Context, options Options) error {
...
 	// So we add them to the default ignore list. See:
 	// https://github.com/GoogleContainerTools/kaniko/blob/63be4990ca5a60bdf06ddc4d10aa4eca0c0bc714/cmd/executor/cmd/root.go#L136
-	ignorePaths := []string{MagicDir, options.LayerCacheDir, options.WorkspaceFolder}
+	ignorePaths := []string{
+		MagicDir,
+		options.LayerCacheDir,
+		options.WorkspaceFolder,
+		// Required for devcontainer to ignore directories like /var/run/secrets,
+		// which frequently has mounted read-only runtime secrets.
+		"/var/run",
```
aaronlehmann, line 480: Does this mean files added to `/var/run` won't be snapshotted? I think that's probably okay in most cases, but might be confusing if someone actually wants to build those files into the container image. Maybe a configurable ignore list would be best?

kylecarbs: I'm changing this to `/var/run/secrets` instead to get closer to the actual problem, and it will become a configurable ignore array in the future that allows users to improve caching without us promoting a bespoke filesystem format.

kylecarbs: Hmm, seems like I actually can't make it be a nested dir because it attempts to remove `/var/run`. I'm going to make this configurable now to save us some headaches.

kylecarbs: It's now configurable! Thanks for the feedback :partying_face:

aaronlehmann: LGTM, thanks

PR #39, opened 2023-07-22
feat: Support alternate path for devcontainer.json file

aaronlehmann:
I have a use case where I need to patch the `devcontainer.json` file that exists in the repo before running the build. This adds a `DEVCONTAINER_JSON_PATH` environment variable which can point to a different path to the `devcontainer.json` file. This allows an alternative `devcontainer.json` to be mounted into the container.

I'm open to suggestions of other ways to do this if anything comes to mind.


kylecarbs (approved):
Reasonable!

PR #40, opened 2023-07-24
feat: Make cache TTL configurable

aaronlehmann:
Add a `CACHE_TTL_DAYS` env var to control layer cache TTL.

Also, fix int handling in `OptionsFromEnv` (which is also needed for `GIT_CLONE_DEPTH`).


envbuilder.go
```diff
@@ -111,6 +111,10 @@ type Options struct {
 	// It will override CacheRepo if both are specified.
 	LayerCacheDir string `env:"LAYER_CACHE_DIR"`
 
+	// CacheTTLDays is the number of days to use cached layers before
+	// expiring them. Defaults to 7 days.
+	CacheTTLDays int `env:"CACHE_TTL_DAYS"`
```
kylecarbs, line 116: Should we make this a `time.Duration` instead? Go can parse pretty nice strings like "7d" or "12hr".

aaronlehmann: I was thinking about that, but since days aren't supported in the format, it's an awkward way to specify long time periods. Okay with me if you prefer that.

kylecarbs (approved):

kylecarbs:
@aaronlehmann could you resolve the new conflicts on your branch? Then feel free to auto-merge!

PR #45, opened 2023-07-26
fix: consume dockerfile arguments when parsing the image

### LINKED ISSUE #35: Dockerfile when using ARG in FROM doesn't work
I was testing the project using this devcontainer specification and envbuilder is failing to build the following Dockerfile => https://github.com/crunchloop/rails-interview/blob/main/.devcontainer/Dockerfile#L3

It throws and error indicating that it can't parse the image. This is working fine with devcontainer cli. I'm going to post the error soon since I don't have it with me.

Removing the `VARIANT` and hard coding the value made it work.

kylecarbs:
Fixes #35.

### DIFF
--- a/devcontainer/devcontainer.go
+++ b/devcontainer/devcontainer.go
@@ -34,7 +34,7 @@ type Spec struct {
 	RemoteUser string            `json:"remoteUser"`
 	RemoteEnv  map[string]string `json:"remoteEnv"`
 	// Features is a map of feature names to feature configurations.
-	Features map[string]map[string]any `json:"features"`
+	Features map[string]any `json:"features"`
 
 	// Deprecated but still frequently used...
 	Dockerfile string `json:"dockerFile"`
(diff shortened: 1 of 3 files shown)


coadler (approved):

devcontainer/devcontainer.go
```diff
@@ -258,21 +258,34 @@ func UserFromDockerfile(dockerfileContent string) string {
...
 // and returns the image that will be used to run the container.
 func ImageFromDockerfile(dockerfileContent string) (name.Reference, error) {
+	args := map[string]string{}
+	var imageRef string
 	lines := strings.Split(dockerfileContent, "\n")
 	// Iterate over lines in reverse
 	for i := len(lines) - 1; i >= 0; i-- {
 		line := lines[i]
-		if !strings.HasPrefix(line, "FROM ") {
+		if strings.HasPrefix(line, "ARG ") {
```
aaronlehmann, line 267: It seems like technically leading whitespace is allowed. See "Note on whitespace" on https://docs.docker.com/engine/reference/builder/.

I'm not sure if multiple spaces (or tabs?) after `ARG` or `FROM` are permitted... but in case they are, tokenizing on whitespace might be a better way to do this.

PR #51, opened 2023-08-16
fix: allow envbuilder on sysbox container runtime

### LINKED ISSUE #50: Envbuilder does not run in a sysbox container
Envbuilder fails if it runs using the sysbox container runtime. This is unfortunate because sysbox is neccessary to run k3s or docker within a workspace.

The issue is described already in the kaniko project: https://github.com/GoogleContainerTools/kaniko/issues/2144 and the sysbox project: https://github.com/nestybox/sysbox/issues/564

There are also workarounds proposed (like unmounting the offending mountpoint or mounting something empty over it). As envbuilder uses kaniko internally, but the image lacks the userland for the workarounds I think, it should also implement the workaround for convenience.

Another option is to attach an empty volume from the outside to the container, but this is difficult to implement in a terraform manifest as the target of the mountpoint is dependend on the running kernel of the host - and the provisioner does not have this information available.

janLo:
This fixes #50 by temporary bind-mounting all readonly mounts within the MagicDir to keep them out of the way for kaniko.

After kaniko finished it's build, the original mountpoints are restored at their original location.

### DIFF
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -523,53 +523,53 @@ func Run(ctx context.Context, options Options) error {
 	mountInfos, err := procfs.GetMounts()
 	if err != nil {
 		logf(codersdk.LogLevelError, "Failed to read mountinfo: %s", err)
-	} else {
-		prefixes := []string{filepath.Join("/", MagicDir), "/proc", "/sys"}
-		for _, mountInfo := range mountInfos {
-			logf(codersdk.LogLevelTrace, "Found mountpoint %s", mountInfo.MountPoint)
-			if _, ok := mountInfo.Options["ro"]; !ok {
-				continue
-			}
-			logf(codersdk.LogLevelTrace, "Found ro mountpoint %s ", mountInfo.MountPoint)
+	}
 
-			relevantMountpoint := true
-			for _, prefix := range prefixes {
-				if strings.HasPrefix(mountInfo.MountPoint, prefix) {
-					logf(codersdk.LogLevelTrace, "Mountpoint %s is NOT relevant (prefix: %s)", mountInfo.MountPoint, prefix)
-					relevantMountpoint = false
-					break
-				}
-			}
-			if !relevantMountpoint {
-				continue
-			}
-
-			src := mountInfo.MountPoint
-			tgt := filepath.Join("/", MagicDir, mountInfo.MountPoint)
-
-			logf(codersdk.LogLevelTrace, "Mountpoint %s is relevant", src)
-
-			logf(codersdk.LogLevelTrace, "Remount mountpoint %s to %s", src, tgt)
-
-			err := os.MkdirAll(tgt, 0750)
-			if err != nil {
-				logf(codersdk.LogLevelError, "Could not create temp mountpoint %s for %s: %s", tgt, src, err.Error())
-				continue
-			}
-
-			err = syscall.Mount(src, tgt, "bind", syscall.MS_BIND, "")
-			if err != nil {
-				logf(codersdk.LogLevelError, "Could not bindmount %s to %s: %s", src, tgt, err.Error())
-				continue
-			}
-			err = syscall.Unmount(src, 0)
-			if err != nil {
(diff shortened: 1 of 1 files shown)


janLo:
It probably needs a lot of polishing as I'm not native to the go ecosystem. I'm sorry about that. But I can confirm that it allows to successfully use envbuilder on top of a sysbox runtime.

janLo:
To give a bit more of an explanation here:

The base-image I'm using in the envbuilder-dockerfile (debian:bookworm) contains a symlink from `/lib` to `/usr/lib/` and kaniko tries to create this. To make way for this symlink, kaniko tries to remove `/lib` in the image - but `lib` is there because sysbox mounts `/lib/modules/<kernelversion>` read-only in the container. Removing the `ro` mount-point fails, of course.

I first attempted to add all `ro` mount points to the ignore list of kaniko - but this means that `/lib/modules/<kernelversion>` is on the ignore-list and this does not match the removal of `/lib` - neither with exact match nor as a prefix. I also think that adding `/lib` to the ignore list would be somewhat invasive.

So my "fix" is now to bind-mount all `ro` mounts which are not in `/proc`,  `/sys` or the `MagicDir` into the `MagicDir` and unmount them at their original locations. After kaniko finished, I bind-mount them back in their original place. I considered to "just unmount" them - but I thought, that there was a reason for them and I couldn't oversee the implications.

**One question that remains** is, if this scheme should also apply to the RW mounts - as this would then be more close to how a "normal" container works (runtime-mounts are "on top" of the image-state).

janLo:
:raised_hand_with_fingers_splayed:

janLo:
:raised_hands:

envbuilder.go
```diff
@@ -516,6 +517,61 @@ func Run(ctx context.Context, options Options) error {
...
+
+		err = syscall.Mount(src, tgt, "bind", syscall.MS_BIND, "")
+		if err != nil {
+			logf(codersdk.LogLevelError, "Could not bindmount %s to %s: %s", src, tgt, err.Error())
+			continue
+		}
+		err = syscall.Unmount(src, 0)
+		if err != nil {
+			logf(codersdk.LogLevelError, "Could not unmount %s: %s", src, err.Error())
+			continue
```
mafredri, line 569: Do we really want to continue in both of these cases? Wouldn't it be safer to error out?

mafredri (commented):
Neat trick, TIL you can unmount the source of a bind-mount. 😄

envbuilder.go
janLo: Looking at it now I think you might be right. On the other hand, the error will catch you later anyway when kaniko tries to delete the containing directory to make room for the base image.

But I do think fail fast is better. Feel free to make any changes you feel suitable.

janLo:
> Neat trick, TIL you can unmount the source of a bind-mount. 😄

There is stuff you learn in 20 years of Linux that you wish you would have never be exposed to 🫣

What do you think about RW mounts? It would mean that we could mount a persistent home directory In an envbuilder container without having to meddle with the ignorelist.

johnstcn:
Creating a separate branch for resolving conflicts: https://github.com/coder/envbuilder/compare/cj/janl/sysbox-issues

Validated using [sysbox v0.6.4 in a Lima VM](https://gist.github.com/johnstcn/b9d6854b3aedb551af1a7588e2a5af94) using envbuilder's own devcontainer:

```
docker run -it --rm \
    -v /tmp/envbuilder:/workspaces \
    -e GIT_URL=https://github.com/coder/envbuilder \
    -e INIT_SCRIPT=bash \
    envbuilder:janl-sysbox-issues # local image
```

Note: dockerd does need to be started manually in the absence of any init or systemd. Some separate fiddling may be necessary there.

johnstcn:
Continuing in https://github.com/coder/envbuilder/pull/183

johnstcn:
Closing this out as #183 has been merged. 

https://github.com/coder/envbuilder/commit/b8e294744e2332572e08c34a5b61158f2cc93521

Thanks for the contribution @janLo !

PR #53, opened 2023-10-02
feat: allow exiting on build failure with fallback

kylecarbs:
This adds support for exiting envbuilder on build failure even when a `FALLBACK_IMAGE` is specified.

### DIFF
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -132,9 +132,11 @@ type Options struct {
 	// container registries.
 	DockerConfigBase64 string `env:"DOCKER_CONFIG_BASE64"`
 
-	// FallbackImage is the image to use if no image is
-	// specified in the devcontainer.json file and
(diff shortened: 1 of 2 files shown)


envbuilder.go
```diff
@@ -137,6 +137,12 @@ type Options struct {
 	// a Dockerfile is not found.
```
aaronlehmann, line 137: Should we update this to mention it also falls back on build failure unless `ExitOnBuildFailure` is set?

aaronlehmann (approved):
Thanks!

envbuilder.go
kylecarbs: Good point, will change!

kylecarbs: Updated!

aaronlehmann: LGTM

PR #55, opened 2023-10-03
Add EXPORT_ENV_FILE parameter

aaronlehmann:
This allows environment variables from the Dockerfile and devcontainer.json to be exported to a file that can be sourced later.

### DIFF
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -676,6 +676,11 @@ func Run(ctx context.Context, options Options) error {
 	_ = file.Close()
 
 	var exportEnvFile *os.File
+	// Do not export env if we skipped a rebuild, because ENV directives
+	// from the Dockerfile would not have been processed and we'd miss these
+	// in the export. We should have generated a complete set of environment
+	// on the intial build, so exporting environment variables a second time
+	// isn't useful anyway.
 	if options.ExportEnvFile != "" && !skippedRebuild {
 		exportEnvFile, err = os.Create(options.ExportEnvFile)
 		if err != nil {
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -194,6 +194,11 @@ type Options struct {
 	// This is useful for self-signed certificates.
 	SSLCertBase64 string `env:"SSL_CERT_BASE64"`
 
+	// ExportEnvFile is an optional file path where envbuilder
+	// will dump environment variables from devcontainer.json
+	// and the built container image.
+	ExportEnvFile string `env:"EXPORT_ENV_FILE"`
+
 	// Logger is the logger to use for all operations.
 	Logger func(level codersdk.LogLevel, format string, args ...interface{})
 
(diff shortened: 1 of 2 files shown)


envbuilder.go
```diff
@@ -724,10 +744,16 @@ func Run(ctx context.Context, options Options) error {
 	for _, env := range configFile.Config.Env {
 		pair := strings.SplitN(env, "=", 2)
 		os.Setenv(pair[0], pair[1])
+		exportEnv(pair[0], pair[1])
```
aaronlehmann, line 747: Environment variables declared with `ENV` in the Dockerfile don't seem to actually show up in `configFile.Config.Env`. This is very strange and I'm not sure why this isn't working as expected.

aaronlehmann: Weird, it works in the test but isn't working in my actual setup 🤔. Trying to dig into this...

aaronlehmann: Ah, found the problem. Env vars from the Dockerfile get skipped when `SKIP_REBUILD=true` and we are starting an existing container. I will update this to skip writing the env export file in this scenario - we only want to write it when we do the actual build.

envbuilder.go
```diff
@@ -668,6 +675,20 @@ func Run(ctx context.Context, options Options) error {
 	}
 	_ = file.Close()
 
+	var exportEnvFile *os.File
+	if options.ExportEnvFile != "" && !skippedRebuild {
```
kylecarbs, line 679: A mention as to why we don't want to do this on rebuild would be helpful. Essentially your GitHub comments!

envbuilder.go
```diff
@@ -194,6 +194,11 @@ type Options struct {
 	// This is useful for self-signed certificates.
 	SSLCertBase64 string `env:"SSL_CERT_BASE64"`
 
+	// ExportEnvFile is an optional file path where envbuilder
+	// will dump environment variables from devcontainer.json
+	// and the built container image.
+	ExportEnvFile string `env:"EXPORT_ENV_FILE"`
```
kylecarbs, line 200: A comment describing the use case would also be helpful, mainly because of the `export` directive, which is technically `sh` syntax, not `.env` file.

kylecarbs: I wonder if we should make this a proper `.env` file being exported instead. Would that make it less useful?

integration/integration_test.go
```diff
@@ -353,6 +353,39 @@ func TestExitBuildOnFailure(t *testing.T) {
 	require.ErrorContains(t, err, "parsing dockerfile")
 }
 
+func TestExportEnvFile(t *testing.T) {
```
kylecarbs, line 356: Great test! 😍

aaronlehmann:
Comments addressed; this should be ready to go :)

kylecarbs (approved):
LGTM!

Issue #34, opened 2023-07-18
Add support for postCreateCommand


bilby91:
Hello,

I've been waiting for this feature in coder for a long time to I'm really excited about helping test the initial versions.

One thing I noticed today was that the `postCreateCommand` doesn't seem to get executed. Is there any idea on how this should be implemented? I'm happy to contribute with this and get myself familiarized with the codebase.

bpmct:
+1. Would be cool if this ran as something in the startup script

aaronlehmann:
+1 from me as well, this would be very useful.

pbabbott:
This would be great! Or at least mention that its not set up yet...

PR #57, opened 2023-10-15
Add support for lifecycle scripts (postCreateCommand, etc)

aaronlehmann:
Parse and execute the following "...Command" hooks:

* `onCreateCommand`
* `updateContentCommand`
* `postCreateCommand`
* `postStartCommand`

Closes: #34

### DIFF
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -844,6 +844,48 @@ func Run(ctx context.Context, options Options) error {
 		return fmt.Errorf("change directory: %w", err)
 	}
 
+	execOneLifecycleScript := func(s devcontainer.LifecycleScript, name string) error {
+		if s.IsEmpty() {
+			return nil
+		}
+		logf(codersdk.LogLevelInfo, "=== Running %s as the %q user...", name, user.Username)
+		if err := s.Execute(ctx, uid, gid); err != nil {
+			logf(codersdk.LogLevelError, "Failed to run %s: %v", name, err)
+			return err
+		}
+		return nil
+	}
+
+	execLifecycleScripts := func() {
+		if !skippedRebuild {
+			if err := execOneLifecycleScript(scripts.OnCreateCommand, "onCreateCommand"); err != nil {
+				return
+			}
+		}
+		if err := execOneLifecycleScript(scripts.UpdateContentCommand, "updateContentCommand"); err != nil {
+			return
+		}
+		if err := execOneLifecycleScript(scripts.PostCreateCommand, "postCreateCommand"); err != nil {
+			return
+		}
+		// TODO: Technically this is supposed to run "each time the
+		// container is successfully started", so would it be more
+		// correct to pass it to InitCommand via the environment and
+		// rely on InitCommand to execute it?
+		if err := execOneLifecycleScript(scripts.PostStartCommand, "postStartCommand"); err != nil {
+			return
+		}
+	}
+
+	// This is called before the Setuid to TARGET_USER because we want the
+	// lifecycle scripts to run using the default user for the container,
+	// rather than the user specified for running the init command. For
+	// example, TARGET_USER may be set to root in the case where we will
+	// exec systemd as the init command, but that doesn't mean we should
+	// run the lifecycle scripts as root.
+	os.Setenv("HOME", user.HomeDir)
+	execLifecycleScripts()
+
 	// The setup script can specify a custom initialization command
 	// and arguments to run instead of the default shell.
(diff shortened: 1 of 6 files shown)


envbuilder.go
```diff
@@ -932,6 +948,41 @@ func Run(ctx context.Context, options Options) error {
...
+		if err := execOneLifecycleScript(scripts.UpdateContentCommand, "updateContentCommand"); err != nil {
+			return
+		}
+		if err := execOneLifecycleScript(scripts.PostCreateCommand, "postCreateCommand"); err != nil {
+			return
+		}
+		// TODO: Technically this is supposed to run "each time the
+		// container is successfully started", so would it be more
+		// correct to pass it to InitCommand via the environment and
+		// rely on InitCommand to execute it?
```
aaronlehmann, line 978: @kylecarbs: Curious for your thoughts on this. It's probably simplest to just execute the post start command here, but technically it's supposed to run once the container starts up successfully, which means the `InitCommand` should handle it. However, it seems burdensome to require an `InitCommand` to support this. I suppose we could make it an option?

envbuilder.go
```diff
@@ -932,6 +948,41 @@ func Run(ctx context.Context, options Options) error {
 
 	logf(codersdk.LogLevelInfo, "=== Running the init command %s %+v as the %q user...", options.InitCommand, initArgs, user.Username)
 
+	execOneLifecycleScript := func(s devcontainer.LifecycleScript, name string) error {
```
aaronlehmann, line 951: Having thought about this a little more, running these after the `Setuid` isn't ideal, because they will run as `TARGET_USER`, which is the intended user for the init command, but not necessarily relevant to the lifecycle scripts.

I'm going to move these to run before the `Setuid` and use `syscall.ForkExec` so they can specify the default user for the container.

envbuilder.go
kylecarbs: Interesting. I'll noodle on this for a sec.

My intuition says there maybe should be a way to manually invoke the `postStartCommand`, I'm just not sure how.

envbuilder.go
kylecarbs: I think that's idiomatic. You might wanna add that as a comment above that code, just so in the future it never gets adjusted!

aaronlehmann: Done!

envbuilder.go
aaronlehmann: Another alternative to passing through the `postStartCommand` through an env variable could be to create a script somewhere on the filesystem that the init command could execute once it's ready to run the `postStartCommand`.

For example, we could support a `POST_START_COMMAND_PATH` option. If you set `POST_START_COMMAND_PATH=/bin/post-start.sh`, envbuilder would create an executable script at the path `/bin/post-start.sh` that would invoke the post start commands. If `POST_START_COMMAND_PATH` is unset, envbuilder would handle the post-start commands itself prior to execing the init command.

aaronlehmann: @kylecarbs: Does this sound like a good approach to you?

kylecarbs: What shell does the official devcontainer spec execute with? That's an oddity I'm not exactly sure how to fix.

I think that route is great though. I primarily want users to be able to execute the post-start command after their own start initialization, just in case they need to prepare something.

aaronlehmann: > What shell does the official devcontainer spec execute with?

The spec says it uses `/bin/sh` to exec commands provided as strings:

> For each command property, if the value is a single string, it will be run in /bin/sh.

aaronlehmann: > I think that route is great though. I primarily want users to be able to execute the post-start command after their own start initialization, just in case they need to prepare something.

Okay, went ahead with this. I added a `POST_START_SCRIPT_PATH` option which will create a shell script at the specified path when set, if any `postStartCommand`s were set. This allows the execution of these commands to be controlled by the init command, so it can do it after successful startup.

kylecarbs: Neato. It's reasonable for us to expect the same then.

It gets weird when the command is an array. What should we do in that scenario?

aaronlehmann: > It gets weird when the command is an array. What should we do in that scenario?

I believe I implemented the proper quoting/escaping to handle this. Every arg is enclosed in single-quotes.

kylecarbs (approved):
So so so great to support this!

kylecarbs:

aaronlehmann:
Thanks for the review!

PR #63, opened 2023-11-02
feature: support dev container in custom subfolder

kconley-sq:
This PR enables the third bullet point described here: https://containers.dev/implementors/spec/#devcontainerjson

> Products using it should expect to find a devcontainer.json file in one or more of the following locations (in order of precedence):
> * `.devcontainer/devcontainer.json`
> * `.devcontainer.json`
> * `.devcontainer/<folder>/devcontainer.json` (where `<folder>` is a sub-folder, one level deep)

Although there is already a `DEVCONTAINER_JSON_PATH` that can be used to configure a custom path to `devcontainer.json` relative to the root `.devcontainer` folder of a repo (e.g. `.devcontainer/my_custom_subfolder/devcontainer.json`), that option can't currently be used with a `Dockerfile` that would be located in that custom path because `envbuilder` still tries to compile the dev container using `.devcontainer` instead of the custom subfolder containing `devcontainer.json`, and meanwhile the `build.Dockerfile` property within `devcontainer.json` [expects](https://containers.dev/implementors/json_reference/#image-specific:~:text=The%20path%20is%20relative%20to%20the%20devcontainer.json%20file.) the provided `Dockerfile` path to be relative to the `devcontainer.json` file which may not be directly under `.devcontainer` if `DEVCONTAINER_JSON_PATH` is configured in this way, so an error occurs like:

```
error: compile devcontainer.json: open dockerfile "/workspaces/.devcontainer/Dockerfile": open /workspaces/.devcontainer/Dockerfile: no such file or directory
```

Instead, this PR introduces a new `DEVCONTAINER_DIR` option that can be used to override the default `.devcontainer` folder to support this use case.

### DIFF
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -407,6 +407,7 @@ func Run(ctx context.Context, options Options) error {
 		if !filepath.IsAbs(devcontainerPath) {
 			devcontainerPath = filepath.Join(devcontainerDir, devcontainerPath)
 		}
+		devcontainerDir = filepath.Dir(devcontainerPath)
 		_, err := options.Filesystem.Stat(devcontainerPath)
 		if err == nil {
 			// We know a devcontainer exists.
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -112,10 +112,16 @@ type Options struct {
 	// It will override CacheRepo if both are specified.
 	LayerCacheDir string `env:"LAYER_CACHE_DIR"`
 
-	// DevcontainerJSONPath is a relative or absolute path to a
-	// devcontainer.json file. This can be used in cases where
-	// one wants to substitute an edited devcontainer.json file
-	// for the one that exists in the repo.
+	// DevcontainerDir is a relative path to the folder containing
+	// the devcontainer.json file that will be used to build the
+	// workspace. If not provided, defaults to `.devcontainer`.
+	DevcontainerDir string `env:"DEVCONTAINER_DIR"`
+
+	// DevcontainerJSONPath is a path to a devcontainer.json file
+	// that is either an absolute path or a path relative to
+	// DevcontainerDir. This can be used in cases where one wants
+	// to substitute an edited devcontainer.json file for the one
+	// that exists in the repo.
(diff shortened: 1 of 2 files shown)


aaronlehmann:
I use `DEVCONTAINER_JSON_PATH` to override the `devcontainer.json` file from the repo (with a patched version), but do not replace the other files in `.devcontainer`. As currently written, this PR would break my use case.

Maybe we could introduce a `DEVCONTAINER_DIR` option and you could set that instead?

kconley-sq:
> I use `DEVCONTAINER_JSON_PATH` to override the `devcontainer.json` file from the repo (with a patched version), but do not replace the other files in `.devcontainer`. As currently written, this PR would break my use case.

I see! I missed that use case was captured in the comment for that option:

```
This can be used in cases where one wants to substitute an edited devcontainer.json file for the one that exists in the repo.
```

> Maybe we could introduce a `DEVCONTAINER_DIR` option and you could set that instead?

Sounds good to me!

kconley-sq:
I'm not sure why but this repo's GitHub Actions PR workflows don't seem to run for my PRs 😞. I retested the latest changes on my fork, though: https://github.com/kconley-sq/envbuilder/actions/runs/6727228526/job/18284743541?pr=1

envbuilder.go
```diff
@@ -112,10 +112,16 @@ type Options struct {
 	// It will override CacheRepo if both are specified.
 	LayerCacheDir string `env:"LAYER_CACHE_DIR"`
 
-	// DevcontainerJSONPath is a relative or absolute path to a
-	// devcontainer.json file. This can be used in cases where
-	// one wants to substitute an edited devcontainer.json file
-	// for the one that exists in the repo.
+	// DevcontainerDir is a relative path to the folder containing
```
aaronlehmann, line 115: Any reason this can't accept an absolute path as well? Supporting both would make it consistent with `DEVCONTAINER_JSON_PATH`.

kylecarbs (approved):
Seems reasonable to me.

Thanks for adding a test too!

PR #65, opened 2023-11-02
fix: using multiple features and opts with spaces

kconley-sq:
https://github.com/coder/envbuilder/pull/64 broke building more than one feature in a dev container due to relative paths provided to `WORKDIR` causing the directory to be changed relative to previous `WORKDIR` directives (e.g. via previous features when multiple features are used), and meanwhile the `MagicDir` was being set to be a relative path. This PR fixes that issue by changing `MagicDir` to be an absolute path.

This PR also fixes an issue around not quoting feature option values which caused building dev containers that use feature option values containing spaces to fail.

An existing integration test around building a dev container with a feature was enhanced to exercise building multiple features that use spaces in their option values.


kylecarbs (approved):
Looks good! Feel free to press merge whenever.

PR #67, opened 2023-11-08
feat: ensure _REMOTE_USER is set for feature install

coryb:
The spec says we should be setting _REMOTE_USER, _CONTAINER_USER, _REMOTE_USER_HOME, and _CONTAINER_USER_HOME.
https://containers.dev/implementors/features/#user-env-var

The _REMOTE_USER is pretty easy for now, not sure about the other 3.

### DIFF
--- a/devcontainer/devcontainer.go
+++ b/devcontainer/devcontainer.go
@@ -29,10 +29,11 @@ func Parse(content []byte) (*Spec, error) {
 }
 
 type Spec struct {
-	Image      string            `json:"image"`
-	Build      BuildSpec         `json:"build"`
-	RemoteUser string            `json:"remoteUser"`
-	RemoteEnv  map[string]string `json:"remoteEnv"`
+	Image         string            `json:"image"`
+	Build         BuildSpec         `json:"build"`
+	RemoteUser    string            `json:"remoteUser"`
+	ContainerUser string            `json:"containerUser"`
+	RemoteEnv     map[string]string `json:"remoteEnv"`
 	// Features is a map of feature names to feature configurations.
 	Features map[string]any `json:"features"`
 	LifecycleScripts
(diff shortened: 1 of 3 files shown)


devcontainer/features/features.go
```diff
@@ -169,8 +169,12 @@ type Spec struct {
 
 // Extract unpacks the feature from the image and returns a set of lines
 // that should be appended to a Dockerfile to install the feature.
-func (s *Spec) Compile(options map[string]any) (string, error) {
-	var runDirective []string
+func (s *Spec) Compile(remoteUser string, options map[string]any) (string, error) {
+	// TODO figure out containerUser so we can set _CONTAINER_USER
```
aaronlehmann, line 173: `containerUser` shouldn't be hard to determine - see [this code](https://github.com/coder/envbuilder/blob/cbbf185544e0104cce4b9c79078cebcdf33266d6/devcontainer/devcontainer.go#L151-L166).

The spec says it can also come from the `containerUser` property in `devcontainer.json`, which we aren't currently parsing out, so we should add that as well:

> The container user can be set with `containerUser` in the `devcontainer.json` and image metadata, `user` in the `docker-compose.yml`, `USER` in the Dockerfile, and can be passed down from the base image.

coryb: Thanks, updated.  I think the logic we had was actually for `containerUser`.  The specs say remoteUser comes from the configs and containerUser comes from configs, USER instructions, or User in the image manifest.  If remoteUser is unset it defaults to containerUser.

aaronlehmann (approved):

devcontainer/devcontainer.go
```diff
@@ -89,7 +90,7 @@ func (s *Spec) Compile(fs billy.Filesystem, devcontainerDir, scratchDir, fallbac
 		env = append(env, key+"="+value)
 	}
 	params := &Compiled{
-		User: s.RemoteUser,
+		User: s.ContainerUser,
```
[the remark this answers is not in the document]
coryb: Correct, my reading of the spec is that we were using containerUser as remoteUser, the difference is pretty minor.

kylecarbs (approved):
Feel free to merge whenever. I'm good with @aaronlehmann's approval :)

PR #72, opened 2023-12-18
Document features missing from devcontainer.json

evilhamsterman:
This is a promising project but there's still some bits missing and it would be helpful to note that for people before they start. I took this list from the commit when it was removed but haven't been following the project that closely, if it is no longer accurate let me know I'll update it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,13 @@
 
 Build development environments from a Dockerfile on Docker, Kubernetes, and OpenShift. Allow developers to modify their environment in a tight feedback loop.
 
-- Supports [`devcontainer.json`](https://containers.dev/) and `Dockerfile`
+- Supports a subset of [`devcontainer.json`](https://containers.dev/) and `Dockerfile`. Currently support is missing for the following `devcontainer.json` features
+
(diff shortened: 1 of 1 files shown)


kylecarbs (approved):

README.md
```diff
@@ -7,7 +7,13 @@
 
 Build development environments from a Dockerfile on Docker, Kubernetes, and OpenShift. Allow developers to modify their environment in a tight feedback loop.
 
-- Supports [`devcontainer.json`](https://containers.dev/) and `Dockerfile`
+- Supports a subset of [`devcontainer.json`](https://containers.dev/) and `Dockerfile`. Currently support is missing for the following `devcontainer.json` features
+
+  - `image`
+  - `build`
+  - `runArgs`
+  - `workspaceFolder`
```
kylecarbs, line 15: We actually support both `image` and `build`!

evilhamsterman: I just pulled it the old commit. Are there things that are there other features that are missing, especially important ones like `customizations` and `mounts`. Looks like lifecycle scripts are supported?

PR #92, opened 2024-02-19
feature: ssh access to git repository

ast9501:
Allow using private key to access repositories.


envbuilder.go
```diff
@@ -41,7 +43,11 @@ import (
 	"github.com/fatih/color"
 	"github.com/go-git/go-billy/v5"
 	"github.com/go-git/go-billy/v5/osfs"
+
+	//"github.com/go-git/go-git/plumbing/transport/ssh"
```
aaronlehmann, line 47: Remove this

go.mod
```diff
@@ -20,6 +20,7 @@ require (
 	github.com/docker/docker v23.0.8+incompatible
 	github.com/fatih/color v1.15.0
 	github.com/go-git/go-billy/v5 v5.5.0
+	github.com/go-git/go-git v4.7.0+incompatible
```
aaronlehmann, line 23: Is this intentional? It seems to be bringing in an older version of go-git alongside v5.

git.go
```diff
@@ -34,20 +33,15 @@ type CloneRepoOptions struct {
...
-		return false, fmt.Errorf("parse url %q: %w", opts.RepoURL, err)
-	}
-	err = opts.Storage.MkdirAll(opts.Path, 0755)
+	var reference string
+
+	err := opts.Storage.MkdirAll(opts.Path, 0755)
 	if err != nil {
 		return false, fmt.Errorf("mkdir %q: %w", opts.Path, err)
 	}
-	reference := parsed.Fragment
```
aaronlehmann, line 45: Removing this seems incorrect?

envbuilder.go
```diff
@@ -194,6 +200,15 @@ type Options struct {
 	// This is optional!
 	GitPassword string `env:"GIT_PASSWORD"`
 
+	// GitSsh will use private key to access Git.
+	GitSsh bool `env:"GIT_SSH"`
```
aaronlehmann, line 204:
```suggestion
	GitSSH bool `env:"GIT_SSH"`
```

envbuilder.go
```diff
@@ -194,6 +200,15 @@ type Options struct {
...
+
+	// GitSshKey is the private key to use for Git authentication.
```
aaronlehmann, line 206:
```suggestion
	// GitSSHKey is the private key to use for Git authentication.
```

envbuilder.go
```diff
@@ -194,6 +200,15 @@ type Options struct {
...
+	// This is optional! (default to $HOME/.ssh/id_rsa)
+	GitSshkey string `env:"GIT_SSHKEY"`
```
aaronlehmann, line 208:
```suggestion
	GitSSHKey string `env:"GIT_SSH_KEY"`
```

envbuilder.go
```diff
@@ -338,25 +353,45 @@ func Run(ctx context.Context, options Options) error {
...
+		if options.GitSsh {
+			if options.GitSshkey != "" {
+				privateKey, err := LoadPrivateKey(options.GitUsername, options.GitSshkey)
+				if err != nil {
+					logf(codersdk.LogLevelError, "Failed to load private ssh key for Git: %s", err)
+				} else {
+					options.RepoAuth = privateKey
+				}
+			} else {
+				logf(codersdk.LogLevelInfo, "GIT_SSH set to True, load default ssh key from path: %s", "/root/.ssh/id_ed25519")
```
aaronlehmann, line 365: I'm not sure this default path makes sense. The key doesn't need to be ED25519 specifically. Maybe it's better to require a path to the key.

envbuilder.go
```diff
@@ -1164,3 +1199,39 @@ type osfsWithChmod struct {
...
+
+	var publicKey *ssh.PublicKeys
+	if username != "" {
+		publicKey, err = ssh.NewPublicKeys(username, []byte(key), "")
+	} else {
+		publicKey, err = ssh.NewPublicKeys("git", []byte(key), "")
+	}
+
+	if err != nil {
+		log.Fatalf("Failed to create ssh auth publickey: ", err)
```
aaronlehmann, line 1232:
```suggestion
		log.Fatalf("Failed to create ssh auth publickey: %s", err)
```

envbuilder.go
```diff
@@ -1164,3 +1199,39 @@ type osfsWithChmod struct {
...
 }
+
+func LoadPrivateKey(username, path string) (*ssh.PublicKeys, error) {
+	file, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+
+	// Get file size and read into byte slice
```
aaronlehmann, line 1210: `os.ReadFile` is a much simpler way to read a file.

envbuilder.go
```diff
@@ -968,11 +1003,11 @@ func DefaultWorkspaceFolder(repoURL string) (string, error) {
...
 		return "/workspaces/empty", nil
 	}
-	parsed, err := url.Parse(repoURL)
-	if err != nil {
-		return "", err
-	}
-	name := strings.Split(parsed.Path, "/")
+
+	reg := regexp.MustCompile(`.*/`)
+	parsed := reg.ReplaceAllString(repoURL, "")
```
aaronlehmann, line 1008:
```go
	parsed := repoURL
	if _, after, ok := strings.Cut(repoURL, "/"); ok {
		parsed = after
	}
```

ast9501: This is a good way to extract the repository name like `git@github.com:<username>/<repository name>.git` (github repository), but it doesn't work for gerrit repository url like `ssh://<username>@<gerrit-server>:<gerrit-ssh-port>/<repository name>`

ast9501 (commented):
I have added some commits for fix/improve some problems according to comment message.

README.md
```diff
@@ -147,6 +148,41 @@ resource "docker_container" "dev" {
 }
 ```
 
+### PrivateKey-based authentication
+* Prebuild private key in your envbuilder image
```
ggjulio, line 152: Isn't this dangerous if we use a cache registry ?
The layer containing the private key may end up in the registry ?

unValerio:
Is there any other way as of right now to clone a private repository?

README.md
ast9501: My original assumption was that this image may be stored in a private registry, so I didn't take into consideration the potential issue of having a private key in the image layers.

ast9501:
> Is there any other way as of right now to clone a private repository?

Maybe you can reference to README doc, there are some guides to access to different kind of git (also for private repository):
https://github.com/coder/envbuilder?tab=readme-ov-file#git-authentication

PR #94, opened 2024-03-02
Support environment variables defined in containerEnv

aaronlehmann:
envbuilder only looked at `remoteEnv`, but it should also define variables from `containerEnv`. `containerEnv` is recommended over `remoteEnv` by the spec:

> We recommend using containerEnv (over remoteEnv) as much as possible
> since it allows all processes to see the variable and isn’t
> client-specific.

### DIFF
--- a/devcontainer/devcontainer.go
+++ b/devcontainer/devcontainer.go
@@ -33,6 +33,7 @@ type Spec struct {
 	Build         BuildSpec         `json:"build"`
 	RemoteUser    string            `json:"remoteUser"`
 	ContainerUser string            `json:"containerUser"`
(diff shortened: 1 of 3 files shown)


kylecarbs (approved):

aaronlehmann:
@kylecarbs: Updated to fix the test. Does GitHub no longer run CI on PRs for this repo, or is there some manual step needed?

aaronlehmann:
Weird, CI ran on this PR: https://github.com/coder/envbuilder/pull/95

aaronlehmann:
It must be because this PR comes from my fork and the other one doesn't.

aaronlehmann:
cc @coryb

PR #96, opened 2024-03-02
Support local features provided inside .devcontainer dir

aaronlehmann:
This allows features to be provided inside the `.devcontainer` directory rather than downloaded from a registry, matching the behavior of the official devcontainer tooling.

For example, Kubernetes tooling generates a devcontainer.json containing:

    "features": {
        "ghcr.io/devcontainers/features/docker-outside-of-docker:1": {
            "enableNonRootDocker": "true",
            "moby": "true"
        },
        "ghcr.io/devcontainers/features/kubectl-helm-minikube:1": {},
        "./local-features/copy-kube-config": {}
    },

cc @coryb


kylecarbs (approved):
Awesome add!

PR #97, opened 2024-03-02
Update deps to support Go 1.21

aaronlehmann:
The dependencies wouldn't build on recent Go versions, which was kind of annoying. Upgrade dependencies so we can use a more recent toolchain.

cc @coryb


aaronlehmann:
Turns out the coder `gvisor` fork doesn't support Go 1.22, but at least it will work with Go 1.21, so we can make some progress here. When Coder upgrades to Go 1.22, that should unblock the upgrade here.

kylecarbs (approved):

kylecarbs:

PR #103, opened 2024-03-08
Fix Dockerfile ARG parsing

### LINKED ISSUE #78: Dockerfile ARG replacement doesnt remove quotes from strings
If you build a Dockerfile in your `devcontainer.json` that includes the following code:
```
ARG VARIANT="3.10"
FROM mcr.microsoft.com/devcontainers/python:0-${VARIANT}
```
or something similar where a string is used with quotes then you get the following error:
`error: compile devcontainer.json: parse image from dockerfile: parse image ref "mcr.microsoft.com/devcontainers/python:0-${VARIANT}\r": could not parse reference: mcr.microsoft.com/devcontainers/python:0-"3.10"`

aaronlehmann:
Double quotes weren't handled while evaluating ARG statements in the Dockerfile.

Fixes #78

### DIFF
--- a/devcontainer/devcontainer.go
+++ b/devcontainer/devcontainer.go
@@ -15,6 +15,7 @@ import (
 	"github.com/go-git/go-billy/v5"
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/moby/buildkit/frontend/dockerfile/shell"
 	"github.com/tailscale/hujson"
 )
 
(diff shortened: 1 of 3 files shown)


devcontainer/devcontainer.go
```diff
@@ -317,30 +318,43 @@ func UserFromDockerfile(dockerfileContent string) string {
...
+				}
+				args = append(args, key+"="+val)
 			}
 			continue
 		}
-		if imageRef == "" && strings.HasPrefix(line, "FROM ") {
-			imageRef = strings.TrimPrefix(line, "FROM ")
+		if imageRef == "" {
+			if fromArgs, ok := strings.CutPrefix(line, "FROM "); ok {
+				imageRef = fromArgs
```
coryb, line 346: can we `strings.TrimSpace(fromArgs)`, I saw another issue with `FROM image<space>` being mis-parsed.

coryb: oh, n/m I see it in your tests, looks good

coryb (approved):

devcontainer/devcontainer.go
aaronlehmann: It's trimmed on this line below:

```go
	image, err := name.ParseReference(strings.TrimSpace(imageRef))
```

I also added a unit test case for this.

kylecarbs (approved):

PR #106, opened 2024-03-13
feat: add proxyoptions to git

michaelbrewer:


envbuilder.go
```diff
@@ -196,6 +197,10 @@ type Options struct {
 	// This is optional!
 	GitPassword string `env:"GIT_PASSWORD"`
 
+	// GitHttpProxyURL is the url for the http proxy.
+	// This is optional!
+	GitHttpProxyURL string `env:"GIT_HTTP_PROXY_URL"`
```
aaronlehmann, line 202: Minor style suggestion:

```suggestion
	GitHTTPProxyURL string `env:"GIT_HTTP_PROXY_URL"`
```

Initialisms in go should be all-caps: https://go.dev/wiki/CodeReviewComments#initialisms

envbuilder.go
```diff
@@ -363,6 +368,17 @@ func Run(ctx context.Context, options Options) error {
 				Password: options.GitPassword,
 			}
 		}
+		if options.GitHttpProxyURL != "" {
+			gitHttpProxyURL, err := url.Parse(options.GitHttpProxyURL)
+			if err != nil {
+				return fmt.Errorf("parse git http proxy url: %w", err)
+			}
+			options.GitHttpProxyURL = gitHttpProxyURL.String()
```
aaronlehmann, line 376: I'm confused about why this value is round-tripped through `URL.Parse`. Why parse it into components and then immediately convert back to a string?

michaelbrewer: Just for validation, but will remove it.

aaronlehmann: Thanks. I think it's fine to validate, but you could discard the result of parsing and only validate that it returns no error:

```go
if _, err := url.Parse(options.GitHttpProxyURL); err != nil {
	return fmt.Errorf("invalid git http proxy url: %w", err)
}
```

Doesn't seem necessary, though. What you have now is fine.

aaronlehmann (approved):

PR #114, opened 2024-03-26
Add registry mirror environment variable option

jw910731:
Add `KANIKO_REGISTRY_MIRROR` environment variable that can set mirrors for docker hub. 
Useful when rate limited by docker hub.

### DIFF
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -656,6 +656,7 @@ func Run(ctx context.Context, options Options) error {
 				Insecure:      options.Insecure,
 				InsecurePull:  options.Insecure,
 				SkipTLSVerify: options.Insecure,
+				RegistryMirrors: strings.Split(os.Getenv("KANIKO_REGISTRY_MIRROR"), ";"),
 			},
 			SrcContext: buildParams.BuildContext,
 		})
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -625,9 +625,9 @@ func Run(ctx context.Context, options Options) error {
 
 		endStage := startStage("🏗️ Building image...")
 		// At this point we have all the context, we can now build!
-		var registry_mirror []string = nil
+		registryMirror := []string{}
 		if val, ok := os.LookupEnv("KANIKO_REGISTRY_MIRROR"); ok {
-			registry_mirror = strings.Split(val, ";")
+			registryMirror = strings.Split(val, ";")
 		}
 		image, err := executor.DoBuild(&config.KanikoOptions{
 			// Boilerplate!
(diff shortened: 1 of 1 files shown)


envbuilder.go
```diff
@@ -656,6 +656,7 @@ func Run(ctx context.Context, options Options) error {
 				Insecure:      options.Insecure,
 				InsecurePull:  options.Insecure,
 				SkipTLSVerify: options.Insecure,
+				RegistryMirrors: strings.Split(os.Getenv("KANIKO_REGISTRY_MIRROR"), ";"),
```
kylecarbs, line 659: What do you think about splitting on ` ` instead? `;` seems a bit less standard to me.

It'd be great to add a comment linking to this PR and [these docs](https://github.com/GoogleContainerTools/kaniko?tab=readme-ov-file#flag---registry-mirror).

kylecarbs (approved):
Thanks for the contribution!

envbuilder.go
jw910731: I references the environment variable design in Kaniko [(link)](https://github.com/GoogleContainerTools/kaniko?tab=readme-ov-file#flag---registry-map), it uses `;` to separate multiple registry map. 
I think it might be better if we follows Kaniko's environment variable format.

kylecarbs: Agreed!

envbuilder.go
```diff
@@ -656,7 +656,11 @@ func Run(ctx context.Context, options Options) error {
 				Insecure:      options.Insecure,
 				InsecurePull:  options.Insecure,
 				SkipTLSVerify: options.Insecure,
-				RegistryMirrors: strings.Split(os.Getenv("KANIKO_REGISTRY_MIRROR"), ";"),
+				// Enables registry mirror features in Kaniko, see more in link below
+				// https://github.com/GoogleContainerTools/kaniko?tab=readme-ov-file#flag---registry-mirror
+				// Related to PR #114
+				// https://github.com/coder/envbuilder/pull/114
+				RegistryMirrors: strings.Split(os.Getenv("·"), ";"),
```
kylecarbs, line 663: I think changing the env var here was a typo!

kylecarbs (approved):

envbuilder.go
aaronlehmann: If the environment variable is unset, this will return `[""]`. Will that give the correct behavior? I would have thought we want to leave this `nil` when there are no mirrors.

jw910731: Thank you for pointing out that I forgot to well consider the case when the environment variable is not set.
It's now fixed. :)

envbuilder.go
```diff
@@ -625,6 +625,10 @@ func Run(ctx context.Context, options Options) error {
 
 		endStage := startStage("🏗️ Building image...")
 		// At this point we have all the context, we can now build!
+		var registry_mirror []string = nil
```
kylecarbs, line 628: To idiomatically match our other var structures, let's initialize this as:
```
registryMirror := []string{}
```

kylecarbs (changes requested):

envbuilder.go
jw910731: Fixed.

kylecarbs (approved):

PR #115, opened 2024-03-27
Remove envbuilder dependency from devcontainer_test

aaronlehmann:
`devcontainer_test` imports the top-level envbuilder package, which creates a transitive dependency on kaniko for external code that imports `envbuilder/devcontainer`. Duplicating `envbuilder.MagicDir` within the test allows us to remove this dependency.

### DIFF
--- a/devcontainer/devcontainer_test.go
+++ b/devcontainer/devcontainer_test.go
@@ -1,4 +1,4 @@
-package envbuilder_test
+package devcontainer_test
 
 import (
 	"crypto/md5"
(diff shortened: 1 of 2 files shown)


kylecarbs (changes requested):
Since we only use `envbuilder.MagicDir` in this file, I'd prefer to copy/paste that var into the test file as a const.

I think it's less idiomatic to split the test from the package rather than duplicate some code.

kylecarbs (approved):

PR #117, opened 2024-03-29
Support providing feature directories in build contexts

aaronlehmann:
When running a build within envbuilder, files created on the filesystem by this Go code are available at the same paths within the container build process. However, if external code calls into the `devcontainer` package to produce a Dockerfile to use for a non-kaniko-based build, the resulting Dockerfile needs to mount the feature directories into the relevant RUN stages in order to access the feature install script and associated files.


kylecarbs (approved):
Awesome!

Issue #116, opened 2024-03-29
Does not search all specified devcontainer locations


bpmct:
It seems like it defaults to looking in `.devconatiner/devcontainer.json` for the file, but does not search all 3 accepted locations in the spec: https://containers.dev/implementors/spec/#devcontainerjson

For example, https://github.com/google/gvisor does not work with the default envbuilder repo passed

PR #134, opened 2024-04-22
feat: locate devcontainer.json in multiple places

mtojek:
Fixes: https://github.com/coder/envbuilder/issues/116

This PR extends the logic responsible for locating the `devcontainer.json` manifest to check in [multiple locations](https://containers.dev/implementors/spec/#devcontainerjson).

### DIFF
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -1195,7 +1195,7 @@ func (fs *osfsWithChmod) Chmod(name string, mode os.FileMode) error {
 }
 
 func findDevcontainerJSON(options Options) (string, string) {
-	// 0. Check provided options
+	// 0. Check provided options.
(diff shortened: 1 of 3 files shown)


envbuilder.go
```diff
@@ -1201,3 +1195,70 @@ type osfsWithChmod struct {
 func (fs *osfsWithChmod) Chmod(name string, mode os.FileMode) error {
 	return os.Chmod(name, mode)
 }
+
+func findDevcontainerJSON(options Options) (string, string, error) {
```
johnstcn, line 1199: this looks like a good candidate for a direct unit test, if it does not already exist.

johnstcn (approved):

envbuilder.go
mtojek: fixed 👍

PR #137, opened 2024-04-23
chore(integration): refactor test git server auth to http mw

johnstcn:
This should allow more easily testing different authentication methods


dannykopping (approved):
LGTM, nice

Issue #129, opened 2024-04-18
Option to specify build path for Dockerfile


bpmct:
You can specify a nested dockerfile location with `DOCKERFILE_PATH`, but you can not specify a build path. For example if you set the `DOCKERFILE_PATH` to `./my/Dockerfile` it will still try to copy assets from root.

Related https://github.com/coder/envbuilder/issues/127

PR #139, opened 2024-04-24
feat: allow custom build context to be configured when using `DOCKERFILE_PATH`

dannykopping:
Fixes https://github.com/coder/envbuilder/issues/129

### DIFF
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -139,6 +139,10 @@ type Options struct {
 	// to using a devcontainer that some might find simpler.
 	DockerfilePath string `env:"DOCKERFILE_PATH"`
 
+	// BuildContextPath can be specified when a DockerfilePath is specified outside the base WorkspaceFolder.
+	// This path MUST be a relative path since it will be joined to the WorkspaceFolder path.
+	BuildContextPath string `env:"BUILD_CONTEXT_PATH"`
+
 	// CacheTTLDays is the number of days to use cached layers before
 	// expiring them. Defaults to 7 days.
 	CacheTTLDays int `env:"CACHE_TTL_DAYS"`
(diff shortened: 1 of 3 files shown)


integration/integration_test.go
```diff
@@ -719,6 +719,91 @@ func TestNoMethodFails(t *testing.T) {
...
+				"a/Dockerfile":  dockerfile,
+				"a/" + inclFile: "...",
+			},
+			dockerfilePath:   "a/Dockerfile",
+			buildContextPath: "", // use default
+			expectedErr:      inclFile + ": no such file or directory",
+		},
+		{
+			// Dockerfile & build context are not in the same dir, but inclFile is in the default build context dir;
+			// this should allow inclFile to be copied. This is probably not desirable though?
```
dannykopping, line 761: Reviewer: should we warn when a Dockerfile is used outside the `WorkspaceFolder`, because the build context will likely be wrong?

johnstcn: You could also make an argument for failing the build entirely; but I think warning is fine here.

johnstcn (approved):

envbuilder.go
```diff
@@ -476,7 +480,7 @@ func Run(ctx context.Context, options Options) error {
 			buildParams = &devcontainer.Compiled{
 				DockerfilePath:    dockerfilePath,
 				DockerfileContent: string(content),
-				BuildContext:      options.WorkspaceFolder,
+				BuildContext:      filepath.Join(options.WorkspaceFolder, options.BuildContextPath),
```
mafredri, line 483: Since `WorkspaceFolder` is optional, what's the end-result if it's not provided, will this be correct?

mafredri (approved):
LGTM, just one question.

mtojek (approved):
Awesome!

envbuilder.go
dannykopping: `WorkspaceFolder` will be given a default value if none is provided: https://github.com/coder/envbuilder/blob/0643d173ef0ad63dae5a7ea3697f16891564b378/envbuilder.go#L270-L276

Issue #130, opened 2024-04-18
Autogenerate docs for all envbuilder flags


bpmct:
There are many flags in envbuilder.go that are not documented in the README. We should document this in a README or another place and make it auto-generated

BrunoQuaresma:
@mtojek my plans for this task:
- Create a go script in `/scripts/doc_envbuilder_flags.go` that adds all the flags to the README.
- I'm wondering if tests can be ignored here since it is not a critical flow.
- Add a step in the CI to check if the README is updated with the latest flags.

Do you think these steps make sense?

mtojek:
Sounds good to me! Only few comments:

`/scripts/doc_envbuilder_flags.go`  - maybe go with a dedicated directory in case we add more scripts with multiple files, so `/scripts/docsgen/main.go`

> I'm wondering if tests can be ignored here since it is not a critical flow.

Yeah I would assume so as in this case results, generated docs, could be reviewed manually in PRs.

 > Add a step in the CI to check if the README is updated with the latest flags.

Yes, must have.

BrunoQuaresma:
The current `Options` structure looks like this:

```go
type Options struct {
	// SetupScript is ran as the root user prior to the init script.
	// It is used to configure envbuilder dynamically during the runtime.
	// e.g. specifying whether to start `systemd` or `tiny init` for PID 1.
	SetupScript string `env:"SETUP_SCRIPT"`
}
```

From this current structure I can get the flag name: `SETUP_SCRIPT` and the type: `string` but I can't get the description as it is a comment. At first, I can see two options:

#### 1. Create a tag called `desc` and add it to the option.
```go
type Options struct {
	SetupScript string `env:"SETUP_SCRIPT" desc:"SetupScript is ran as the root user prior to the init script. It is used to configure envbuilder dynamically during the runtime. e.g. specifying whether to start 'systemd' or 'tiny init' for PID 1."`
}
```
This option doesn't require major code changes, but having the description as a single-line string can be difficult to read.

#### 2. Create another type to encapsulate each option
```go
type Option struct {
	Env string
	Description string
	Type interface{}
}

type Options struct {
	SetupScript Option{
		Env: "SETUP_SCRIPT",
		Description:  `
			SetupScript is ran as the root user prior to the init script.
			It is used to configure envbuilder dynamically during the runtime.
			e.g. specifying whether to start 'systemd' or 'tiny init' for PID 1
		`,
		Type: string,
	}
}
```
[...]

PR #140, opened 2024-04-24
refactor: refactor envbuilder to use coder/serpent as CLI engine

BrunoQuaresma:
Related to https://github.com/coder/envbuilder/issues/130


envbuilder.go
```diff
@@ -77,152 +76,9 @@ var (
 	MagicFile = filepath.Join(MagicDir, "built")
 )
 
-type Options struct {
```
mtojek, line 80: I'd like to preserve the structure for Options as it ensures type safety. You don't need to use `GetString()`, `GetBool()` methods as Go can handle this natively.

envbuilder.go
```diff
@@ -233,60 +89,49 @@ type Options struct {
...
-	if options.InitCommand == "" {
-		options.InitCommand = "/bin/sh"
-	}
-	if options.IgnorePaths == nil {
-		// Kubernetes frequently stores secrets in /var/run/secrets, and
-		// other applications might as well. This seems to be a sensible
-		// default, but if that changes, it's simple to adjust.
-		options.IgnorePaths = []string{"/var/run"}
-	}
+func Run(ctx context.Context, options OptionsMap, deps Dependencies) error {
```
mtojek, line 92: The idea of extracting `Logger` and `FileSystem` from `Options` is correct, but it is more natural to just keep them as separate arguments, so:

```go
func Run(ctx context.Context, options OptionsMap, logger Logger, filesystem Filesystem) error {
```

envbuilder.go
```diff
@@ -77,152 +76,9 @@ var (
...
-	ExportEnvFile string `env:"EXPORT_ENV_FILE"`
-
-	// PostStartScriptPath is the path to a script that will be created by
-	// envbuilder based on the `postStartCommand` in devcontainer.json, if any
-	// is specified (otherwise the script is not created). If this is set, the
-	// specified InitCommand should check for the presence of this script and
-	// execute it after successful startup.
-	PostStartScriptPath string `env:"POST_START_SCRIPT_PATH"`
-
+type Dependencies struct {
```
johnstcn, line 79: Why do we need a separate struct here?

mtojek (changes requested):
I think you can review [clidocgen](https://github.com/coder/coder/tree/main/scripts/clidocgen) in coder/coder and try to implement something similar or extract the clidocgen to a common /library repo?

options_test.go
```diff
@@ -0,0 +1,74 @@
+package envbuilder_test
+
+import (
+	"testing"
+
+	"github.com/coder/envbuilder"
+	"github.com/stretchr/testify/require"
+)
+
+func TestOptionsFromEnv(t *testing.T) {
```
johnstcn, line 10: We'd probably want a separate test to validate the defaults with an empty env.

johnstcn (commented):
It probably makes sense to use `coder/serpent` here as what I'm seeing is essentially a reimplementation of a subset of its existing functionality.

envbuilder.go
BrunoQuaresma: Hum.. how I can preserve the `Options` structure and add a description for each item? I talked to @mafredri about having the descriptions on the CLI only but making this we guess developers are not going to keep it updated since it would be in a different place where the option is defined.

envbuilder.go
mtojek: yes, but as separate function arguments, not a dedicated `Deps` wrapper.

envbuilder.go
BrunoQuaresma: Since we want to use `coder/serpent` for this I think the `Options`  will need to change anyway.

options_test.go
BrunoQuaresma: Now, env variables are handled by `coder/serpent` so I don't see a need for us to test it anymore.

cmd/envbuilder/main.go
```diff
@@ -23,15 +23,14 @@ import (
 )
 
 func main() {
-	root := &cobra.Command{
-		Use: "envbuilder",
+	config := envbuilder.Config{}
```
mtojek, line 26: nit: alternatively you can write it as:

`var config envbuilder.Config`

config.go
```diff
@@ -0,0 +1,280 @@
...
+			to substitute an edited devcontainer.json file for the one
+			that exists in the repo.`,
+		},
+		{
+			Name:  "Dockerfile Path",
+			Env:   "DOCKERFILE_PATH",
+			Value: serpent.StringOf(&c.DockerfilePath),
+			Description: `DockerfilePath is a relative path to the Dockerfile that
+			will be used to build the workspace. This is an alternative
+			to using a devcontainer that some might find simpler.`,
```
mtojek, line 126: I think Go will keep the break in the string so this will become: `...to the Dockerfile that			will be used to build the workspace.`

Just use a one-liner `Description: "..."`,

config.go
```diff
@@ -0,0 +1,280 @@
...
+	WorkspaceFolder      string
+	SSLCertBase64        string
+	ExportEnvFile        string
+	PostStartScriptPath  string
+}
+
+func (c *Config) Options() serpent.OptionSet {
+	return serpent.OptionSet{
+		{
+			Name:  "Setup Script",
```
mtojek, line 42: I'm looking at coder/coder. Looks like you can skip `Name` property as it is redundant? See in coder/coder:

```
		{
			Flag:        "script-data-dir",
			Default:     os.TempDir(),
			Description: "Specify the location for storing script data.",
			Env:         "CODER_AGENT_SCRIPT_DATA_DIR",
			Value:       serpent.StringOf(&scriptDataDir),
		},

```

envbuilder.go
```diff
@@ -1209,22 +1006,23 @@ func (fs *osfsWithChmod) Chmod(name string, mode os.FileMode) error {
 	return os.Chmod(name, mode)
 }
 
-func findDevcontainerJSON(options Options) (string, string, error) {
+func findDevcontainerJSON(c *Config, fs billy.Filesystem, logger Logger) (string, string, error) {
```
mtojek, line 1009: I would keep the name `Options` to reduce the blast radius of PR changes. Happy to see more comments on this.

Also, for simple structs (this is a simple struct), it should be `c Config` or (`options Options`)

envbuilder.go
```diff
@@ -1238,35 +1036,34 @@ func findDevcontainerJSON(options Options) (string, string, error) {
...
+	fileInfos, err := fs.ReadDir(devcontainerDir)
 	if err != nil {
 		return "", "", err
 	}
 
-	logf := options.Logger
 	for _, fileInfo := range fileInfos {
 		if !fileInfo.IsDir() {
-			logf(codersdk.LogLevelDebug, `%s is a file`, fileInfo.Name())
+			logger(codersdk.LogLevelDebug, `%s is a file`, fileInfo.Name())
```
mtojek, line 1060: Same here: stick to `logf` and reduce the PR size

cmd/envbuilder/main.go
```diff
@@ -23,15 +23,14 @@ import (
...
+		Use:     "envbuilder",
+		Options: config.Options(),
 		// Hide usage because we don't want to show the
 		// "envbuilder [command] --help" output on error.
-		SilenceUsage:  true,
-		SilenceErrors: true,
-		RunE: func(cmd *cobra.Command, args []string) error {
-			options := envbuilder.OptionsFromEnv(os.LookupEnv)
-
+		Hidden: true,
```
mtojek, line 32: Hidden: false?

Looks like in coder/serpent this property means something else: `// Hidden determines whether the command should be hidden from help.`

config.go
```diff
@@ -0,0 +1,280 @@
+package envbuilder
+
+import (
+	"github.com/coder/serpent"
+)
+
+type Config struct {
```
mtojek, line 7: Convention: It is a good practice to add comments to exposed structure fields (capital letter = exposed = visible outside the package).

See: https://github.com/coder/serpent/blob/00b6c3f9d4bced45b5f35278853c07f0548ba53f/command.go#L25

envbuilder.go
```diff
@@ -428,17 +261,17 @@ func Run(ctx context.Context, options Options) error {
...
-	if options.DockerfilePath == "" {
+	if c.DockerfilePath == "" {
 		// Only look for a devcontainer if a Dockerfile wasn't specified.
 		// devcontainer is a standard, so it's reasonable to be the default.
-		devcontainerPath, devcontainerDir, err := findDevcontainerJSON(options)
+		devcontainerPath, devcontainerDir, err := findDevcontainerJSON(c, fs, logger)
 		if err != nil {
-			logf(codersdk.LogLevelError, "Failed to locate devcontainer.json: %s", err.Error())
-			logf(codersdk.LogLevelError, "Falling back to the default image...")
+			logger(codersdk.LogLevelError, "Failed to locate devcontainer.json: %s", err.Error())
```
mtojek, line 269: Same: logger -> logf

envbuilder_internal_test.go
```diff
@@ -36,10 +33,7 @@ func TestFindDevcontainerJSON(t *testing.T) {
 		require.NoError(t, err)
 
 		// when
-		_, _, err = findDevcontainerJSON(Options{
-			Filesystem:      fs,
-			WorkspaceFolder: "/workspace",
-		})
+		_, _, err = findDevcontainerJSON(&Config{WorkspaceFolder: "/workspace"}, fs, nil)
```
mtojek, line 36: I don't mind configuring a logger instance for testing purposes :)

mtojek (commented):
I believe you are heading in the right direction. Most of my comments pertain to adhering to the Go convention, but the overall concept is correct.

I suggest narrowing the scope of this PR to solely adopting `coder/serpent` (what you have now) and then preparing a follow-up to implement `clidocgen`. What do you think?

envbuilder_test.go
```diff
@@ -17,36 +17,3 @@ func TestDefaultWorkspaceFolder(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, envbuilder.EmptyWorkspaceDir, dir)
 }
-
-func TestSystemOptions(t *testing.T) {
```
johnstcn, line 21: Up to you, but I think it makes sense to keep this around and test:
- That all options have a sensible default value
- That all options can be overridden

cmd/envbuilder/main.go
BrunoQuaresma: What is the difference and why is this preferred?

config.go
BrunoQuaresma: Ahhhh, I see 🤔

config.go
BrunoQuaresma: Ahh I see, so we can just use `Flag` right?

envbuilder.go
BrunoQuaresma: What defines if a struct is complex or simple?

cmd/envbuilder/main.go
BrunoQuaresma: Since this is one command CLI we don't want to show it in the help as an "option" but I'm not sure I'm 100% right. @ammario maybe help us.

envbuilder_internal_test.go
BrunoQuaresma: Not sure what it means...

envbuilder_test.go
BrunoQuaresma: I feel I would be testing Cobra stuff, but I will put more thoughts on that for sure.

cmd/envbuilder/main.go
johnstcn: `var x Thing` is just delcaration https://go.dev/ref/spec#Variable_declarations
`x := Thing` is both declaration and assignment https://go.dev/ref/spec#Short_variable_declarations

IMO it's dealer's choice.

mtojek: I'm fine with leaving as is!

cmd/envbuilder/main.go
ammario: I don't think Hidden has any effect on the root command. It determines whether the command shows up as a subcommand to its parent during help, and there is no parent here.

envbuilder_internal_test.go
mtojek: I meant that you can use a real logger here too:

not sure if exactly this API, but I believe you get the concept :)
```
logger := logrus.New()
logger.SetLevel(log.DebugLevel)
_, _, err = findDevcontainerJSON(&Config{WorkspaceFolder: "/workspace"}, fs, logger)
```

options.go
```diff
@@ -0,0 +1,267 @@
...
+				"file.\n\nSetupScript is ran as the root user prior to the init script. " +
+				"It is used to configure envbuilder dynamically during the runtime. e.g. " +
+				"specifying whether to start systemd or tiny init for PID 1.",
+		},
+		{
+			Flag:        "init-script",
+			Env:         "INIT_SCRIPT",
+			Default:     "sleep infinity",
+			Value:       serpent.StringOf(&o.InitScript),
+			Description: "InitScript is the script to run to initialize the workspace.",
```
dannykopping, line 58: Stylistic nit: prefixing the descriptions with the variable name is unnecessary, and likely a consequence of moving the text from comments (which follow Go's conventions) to here.

`InitScript` isn't a valid CLI option; `init-script` and `INIT_SCRIPT` are.

Example refactor:

```suggestion
			Description: "The script to run to initialize the workspace",
```

envbuilder.go
```diff
@@ -1137,43 +971,6 @@ func createPostStartScript(path string, postStartCommand devcontainer.LifecycleS
 	return nil
 }
 
-// OptionsFromEnv returns a set of options from environment variables.
-func OptionsFromEnv(getEnv func(string) (string, bool)) Options {
```
dannykopping, line 1141: It feels like some subtle bugs could arise from removing this, and relying on serpent's handling of envs. Consider [Hyrum's Law](https://www.hyrumslaw.com/); if we're changing the way envs are handled then it may break someone's usage; I think we should have a set of tests covering this.

dannykopping: This could be in a follow-up PR.

cmd/envbuilder/main.go
```diff
@@ -23,15 +23,11 @@ import (
...
-	root := &cobra.Command{
-		Use: "envbuilder",
-		// Hide usage because we don't want to show the
-		// "envbuilder [command] --help" output on error.
-		SilenceUsage:  true,
-		SilenceErrors: true,
-		RunE: func(cmd *cobra.Command, args []string) error {
-			options := envbuilder.OptionsFromEnv(os.LookupEnv)
-
+	options := envbuilder.Options{}
```
dannykopping, line 26: This is the Go style for zero-value instantiations.

```suggestion
	var options envbuilder.Options
```

options.go
johnstcn: For context, this is what it looks like in the `--help` output:

```
       string, $INIT_SCRIPT (default: sleep infinity)
          InitScript is the script to run to initialize the workspace.
```

As a follow-up, it might make sense to add a test for this against a `.golden` file.

envbuilder.go
johnstcn: I second this; there may be cases that our existing integration tests are not catching.

cmd/envbuilder/main.go
dannykopping: This is a stylistic nit

https://go.dev/ref/spec#The_zero_value
https://go.dev/tour/basics/12

```bash
$ find . -name '*.go' -exec egrep '.*:=\s*\w*\s*\{\s*\}$' -rin . {} \;
./enterprise/tailnet/htmldebug.go:64:   out := HTMLDebug{}
./enterprise/coderd/coderdenttest/coderdenttest.go:131:                 lo := LicenseOptions{}
./tailnet/node_internal_test.go:586:    d := PeerDiagnostics{}
./tailnet/node_internal_test.go:601:            d := PeerDiagnostics{}
./tailnet/node_internal_test.go:634:    d := PeerDiagnostics{}
./agent/agentssh/jetbrainstrack.go:34:  d := localForwardChannelData{}
./coderd/database/dbmem/dbmem.go:3634:  templates := templateRow{}
./coderd/devtunnel/tunnel.go:110:       cfg := Config{}
./coderd/audit/request.go:411:  d := WorkspaceBuildBaggage{}
./scripts/apitypings/main.go:526:       state := structTemplateState{}
./scripts/audittypegen/main.go:32:      g := Generator{}
./codersdk/apikey.go:188:       tokenConfig := TokenConfig{}
./site/site.go:333:     noopRW := noopResponseWriter{}
..
```

We have plenty of cases of this in the coder codebase, so it's not a firm suggestion but more of a "this is how Go encourages usage, but it's pretty-much equivalent either way"

options.go
mafredri: I'd personally like to see golden file generation/validation as a part of serpents design (subpackage?), would be a cool feature for applications that need strict guarantees.

cmd/envbuilder/main.go
mafredri: I would agree with you @dannykopping that `var options envbuilder.Options` is more idiomatic, but I'm mostly basing that on experience over the years. I don't think any of the references mention a preference of one over the other. There's also the Go code review comments document which also doesn't mention it in any way.

https://go.dev/wiki/CodeReviewComments

envbuilder.go
mafredri: Since this is public API, it should probably be made backwards compatible anyway. We may add a deprecation notice on this if we don't want it used anymore:

https://go.dev/wiki/Deprecated

envbuilder.go
```diff
@@ -1204,13 +1001,14 @@ func (fs *osfsWithChmod) Chmod(name string, mode os.FileMode) error {
 	return os.Chmod(name, mode)
 }
 
-func findDevcontainerJSON(options Options) (string, string, error) {
+func findDevcontainerJSON(options Options, fs billy.Filesystem, logf Logger) (string, string, error) {
```
mafredri, line 1004:
```suggestion
func findDevcontainerJSON(fs billy.Filesystem, logf Logger, options Options) (string, string, error) {
```
(See other comment, but ideally we'd move these back into options IMO.)

options.go
```diff
@@ -0,0 +1,267 @@
+package envbuilder
+
+import (
+	"github.com/coder/serpent"
+)
+
+// Options contains the configuration for the envbuilder.
+type Options struct {
+	SetupScript          string
```
mafredri, line 9: I think this change is quite unfortunate as it removes all documentation from Go docs (comment above field).

It'd be nice if we could keep the comments on the struct and instead `go generate` the serpent descriptions from those.

If we don't amend this (comments on fields), it doesn't really make sense to keep this a public/exported struct.

envbuilder.go
johnstcn: Envbuilder is still 0.x so there are no compatibility guarantees right now. If we want to make breaking API changes, now is the time.

mafredri: That's fair 👍🏻.

envbuilder.go
```diff
@@ -77,184 +77,16 @@ var (
...
-	if options.IgnorePaths == nil {
-		// Kubernetes frequently stores secrets in /var/run/secrets, and
-		// other applications might as well. This seems to be a sensible
-		// default, but if that changes, it's simple to adjust.
-		options.IgnorePaths = []string{"/var/run"}
-	}
+// Logger is the logf to use for all operations.
+// Filesystem is the filesystem to use for all operations.
+// Defaults to the host filesystem.
+func Run(ctx context.Context, options Options, fs billy.Filesystem, logf Logger) error {
```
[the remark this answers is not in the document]
mtojek: I suggested this to ensure options have a single responsibility - just configuration.

options.go
mtojek: > It'd be nice if we could keep the comments on the struct and instead go generate the serpent descriptions from those.

I'm happy to see `go generate` adoption here, but maybe in a separate issue/PR? Let's not overwhelm Bruno with extra magical Go spells.

envbuilder.go
```diff
@@ -517,9 +350,9 @@ func Run(ctx context.Context, options Options) error {
 		}
 
 		// Disable all logging from the registry...
-		logger := logrus.New()
-		logger.SetOutput(io.Discard)
-		entry := logrus.NewEntry(logger)
+		logrusLogger := logrus.New()
+		logrusLogger.SetOutput(io.Discard)
+		entry := logrus.NewEntry(logrusLogger)
```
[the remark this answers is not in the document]
BrunoQuaresma: This is interesting. Ty

envbuilder.go
BrunoQuaresma: Ok, since @johnstcn and @dannykopping have brought up the concern about how Serpent handles env vars, I will add a test for it 💯 😆. Would this be good enough or do I need to take any extra action?

envbuilder.go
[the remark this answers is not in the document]
BrunoQuaresma: @mafredri since `fs` and `logf` could be `nil`, would not make sense to have them at last?

envbuilder.go
```diff
@@ -775,7 +608,7 @@ func Run(ctx context.Context, options Options) error {
 
 	// Remove the Docker config secret file!
 	if options.DockerConfigBase64 != "" {
-		err = os.Remove(filepath.Join(MagicDir, "config.json"))
+		err = os.Remove(filepath.Join(MagicDir, "o.json"))
```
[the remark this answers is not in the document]
BrunoQuaresma: It is not, good catch!

envbuilder.go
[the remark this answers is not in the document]
mafredri: > I suggested this to ensure options have a single responsibility - just configuration.

@mtojek Given that they are optional, I see them as options. I don't think it makes sense to limit the concept of options to "can it be passed via env or not", we already define that via serpent.

envbuilder.go
```diff
@@ -356,11 +189,11 @@ func Run(ctx context.Context, options Options) error {
 
 		cloneOpts := CloneRepoOptions{
 			Path:         options.WorkspaceFolder,
-			Storage:      options.Filesystem,
+			Storage:      fs,
 			Insecure:     options.Insecure,
 			Progress:     writer,
 			SingleBranch: options.GitCloneSingleBranch,
-			Depth:        options.GitCloneDepth,
+			Depth:        int(options.GitCloneDepth),
```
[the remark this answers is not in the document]
mafredri: Ah, gotcha. Yeah that's a limitation in serpent that could probably be solved with generics, or at least via expanded API. Thanks for the explanation 👍🏻

envbuilder.go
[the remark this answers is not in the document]
johnstcn: This should have caused the integration tests to fail 🤔

options.go
mafredri: @BrunoQuaresma There's usually just 2 components to golden file testing:

1. Capture output from the command(s) you want to verify
2. Write update, or verify output against existing file (e.g. `testdata/help/envbuilder--help.golden`)

To achieve 1) you set the serpent invocation stdout/stderr output to a buffer (e.g. `bytes.Buffer`) and run it.

To achieve 2) you add `var updateGoldenFiles = flag.Bool("update", false, "update .golden files")` to the test file:
- If true, write the output from 1) to the golden file
- If false, read output from the golden file and do `if diff := cmp.Diff(want, got); diff != "" { require.Failf(t, "mismatch (-want +got):\n%s", diff) }`

There's some precedent in `coder/coder` but I wouldn't recommend copying it directly as there's a lot we don't need from there: https://github.com/coder/coder/blob/main/cli/clitest/golden.go

envbuilder.go
BrunoQuaresma: Marking this as resolved. I'm going to post a comment with all the follow-ups later.

envbuilder.go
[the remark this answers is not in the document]
BrunoQuaresma: I can get them back to leave inside of `Options` for sure.

envbuilder.go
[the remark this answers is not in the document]
BrunoQuaresma: It definitely should 🤔 do you have any idea why it is not panicking?

envbuilder.go
BrunoQuaresma: Going to add the logf and fs back into the Options

options.go
BrunoQuaresma: Marking this as resolved for now. Going to comment what are all the follow-ups later.

envbuilder.go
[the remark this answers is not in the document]
BrunoQuaresma: Should we? Or does it not matter?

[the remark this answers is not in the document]
johnstcn: You can open up a separate issue if you like, but I think it's out of scope of this PR.
At most, we should warn.

BrunoQuaresma:
@mafredri @johnstcn @dannykopping I think this PR is in a good state to get merged. Here are some followups from the reviews:
- Test CLI output against a golden file . [Reference]( https://github.com/coder/envbuilder/pull/140#discussion_r1581071538 )
- Test to validate that envs given previously will still produce the same options now in Serpent . [Reference](https://github.com/coder/envbuilder/pull/140#discussion_r1581024599)
-  Keep the comments on the struct and instead go generate the serpent descriptions from those. [Reference](https://github.com/coder/envbuilder/pull/140#discussion_r1580870391)

I'm planning to create an issue + PR for each one of them after merging https://github.com/coder/envbuilder/pull/147.

envbuilder.go
```diff
@@ -269,25 +99,26 @@ func Run(ctx context.Context, options Options) error {
 	}
 	if options.WorkspaceFolder == "" {
 		var err error
-		options.WorkspaceFolder, err = DefaultWorkspaceFolder(options.GitURL)
+		folder, err := DefaultWorkspaceFolder(options.GitURL)
+		options.WorkspaceFolder = folder
```
johnstcn, line 103: nit: assigning `options.WorkspaceFolder` should probably go after `return err`, and you can probably also remove the declaration of `err` on line 101.

dannykopping (approved):
LGTM

johnstcn (approved):
:shipit:

mafredri (approved):

Issue #126, opened 2024-04-15
Gitlab oauth2:token being written to .git/config file after cloning


ns-mkusper:
I'm using envbuilder `0.2.9` and allowing my users to set the branch portion of the `GIT_URL` (by using a parameter in the `#refs/heads/${data.coder_parameter.repo_branch.value}` value) in new workspaces via the form parameters. This works, but when the repo gets cloned the credentials I set to `GIT_USERNAME` and `GIT_PASSWORD` for envbuilder to be able to pull the repo are written to `.git/config` (as explained [in the readme](https://github.com/coder/envbuilder?tab=readme-ov-file#git-authentication) ) which then overrides any token refreshing as the config takes priority over `GIT_ASKPASS`. 

As a workaround I am just clearing this via a startup script, but this seems to be a bug.

bpmct:
related https://github.com/coder/envbuilder/issues/91

PR #141, opened 2024-04-24
fix: do not inject GIT_USERNAME and GIT_PASSWORD into git clone URL

johnstcn:
Fixes https://github.com/coder/envbuilder/issues/126

We had been writing the `GIT_USERNAME` and `GIT_PASSWORD` into the git auth URL as well as setting basic HTTP auth credentials.
This causes `GIT_USERNAME` and `GIT_PASSWORD` to be present in the `.git/config` of the clond repo, which is not desirable. 
This PR removes that behaviour and modifies existing unit tests to assert that the git clone URL is not modified from what the user passes in.

The previous behaviour can be preserved by explicitly setting basic auth in the git clone URL, if required.

### DIFF
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -361,12 +361,12 @@ func Run(ctx context.Context, options Options) error {
 		}
 
 		if options.GitUsername != "" || options.GitPassword != "" {
-			gitURL, err := url.Parse(options.GitURL)
-			if err != nil {
-				return fmt.Errorf("parse git url: %w", err)
-			}
-			gitURL.User = url.UserPassword(options.GitUsername, options.GitPassword)
-			options.GitURL = gitURL.String()
+			// Previously, we had been placing credentials in the URL
+			// as well as setting githttp.BasicAuth.
+			// This was removed as it would leak the credentials used
+			// to clone the repo into the resulting workspace.
+			// Users may still hard-code credentials directly into the
+			// git URL themselves, if required.
 
 			cloneOpts.RepoAuth = &githttp.BasicAuth{
 				Username: options.GitUsername,
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -361,13 +361,8 @@ func Run(ctx context.Context, options Options) error {
 		}
 
 		if options.GitUsername != "" || options.GitPassword != "" {
-			// Previously, we had been placing credentials in the URL
-			// as well as setting githttp.BasicAuth.
-			// This was removed as it would leak the credentials used
-			// to clone the repo into the resulting workspace.
-			// Users may still hard-code credentials directly into the
-			// git URL themselves, if required.
-
+			// NOTE: we previously inserted the credentials into the repo URL.
+			// This was removed in https://github.com/coder/envbuilder/pull/141
 			cloneOpts.RepoAuth = &githttp.BasicAuth{
 				Username: options.GitUsername,
 				Password: options.GitPassword,
--- a/git_test.go
+++ b/git_test.go
@@ -2,10 +2,12 @@ package envbuilder_test
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"net/http/httptest"
 	"net/url"
 	"os"
+	"regexp"
 	"testing"
 	"time"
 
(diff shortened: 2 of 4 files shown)


integration/integration_test.go
```diff
@@ -756,27 +780,12 @@ type gitServerOptions struct {
...
 	t.Helper()
 	if opts.authMW == nil {
-		opts.authMW = checkBasicAuth(opts.username, opts.password)
+		opts.authMW = gittest.BasicAuthMW(opts.username, opts.password)
 	}
 	srv := httptest.NewServer(opts.authMW(createGitHandler(t, opts)))
 	return srv.URL
 }
 
-func checkBasicAuth(username, password string) func(http.Handler) http.Handler {
```
johnstcn, line 765: review: moved to `gittest`

git_test.go
```diff
@@ -4,72 +4,198 @@ import (
...
+	githttp "github.com/go-git/go-git/v5/plumbing/transport/http"
 	"github.com/stretchr/testify/require"
 )
 
 func TestCloneRepo(t *testing.T) {
 	t.Parallel()
 
-	t.Run("Clones", func(t *testing.T) {
-		t.Parallel()
+	for _, tc := range []struct {
```
johnstcn, line 25: review: trying to cover all cases outlined here: 

https://github.com/coder/envbuilder?tab=readme-ov-file#git-authentication

envbuilder.go
```diff
@@ -361,12 +361,12 @@ func Run(ctx context.Context, options Options) error {
...
-				return fmt.Errorf("parse git url: %w", err)
-			}
-			gitURL.User = url.UserPassword(options.GitUsername, options.GitPassword)
-			options.GitURL = gitURL.String()
+			// Previously, we had been placing credentials in the URL
+			// as well as setting githttp.BasicAuth.
+			// This was removed as it would leak the credentials used
+			// to clone the repo into the resulting workspace.
+			// Users may still hard-code credentials directly into the
+			// git URL themselves, if required.
```
johnstcn, line 369: review: alternatively, we can keep this and munge the git config afterwards.

integration/integration_test.go
```diff
@@ -71,13 +71,37 @@ func TestSucceedsGitAuth(t *testing.T) {
...
+	u, err := url.Parse(gitURL)
+	require.NoError(t, err)
+	u.User = url.UserPassword("kyle", "testing")
+	ctr, err := runEnvbuilder(t, options{env: []string{
+		"GIT_URL=" + u.String(),
+		"DOCKERFILE_PATH=Dockerfile",
+	}})
+	require.NoError(t, err)
+	gitConfig := execContainer(t, ctr, "cat /workspaces/.git/config")
+	require.Contains(t, gitConfig, u.String())
```
johnstcn, line 104: review: Check here is to ensure that we don't modify git clone URL

gittest/gittest.go
```diff
@@ -118,3 +118,18 @@ func WriteFile(t *testing.T, fs billy.Filesystem, path, content string) {
 	err = file.Close()
 	require.NoError(t, err)
 }
+
+func BasicAuthMW(username, password string) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if username != "" && password != "" {
```
dannykopping, line 125: Wouldn't this mean that if you pass either `username` or `password` then auth will be bypassed?

```suggestion
			if username != "" || password != "" {
```

integration/integration_test.go
dannykopping: I'd suggest moving this to a comment.

git_test.go
```diff
@@ -4,72 +4,198 @@ import (
...
+					return
+				}
+				require.NoError(t, err)
+				require.True(t, cloned)
+
+				readme := mustRead(t, clientFS, "/workspace/README.md")
+				require.Equal(t, "Hello, world!", readme)
+				gitConfig := mustRead(t, clientFS, "/workspace/.git/config")
+				// We do not modify the git URL that folks pass in.
+				require.Contains(t, gitConfig, authURL.String())
```
dannykopping, line 163: This may be a bit flimsy; you could harden it with:

```suggestion
				require.Regexp(t, fmt.Sprintf("^\s*url = %s$", regexp.QuoteMeta(authURL.String()), gitConfig)
```

**NOTE**: I have not tested the above code

envbuilder.go
mtojek: Maybe rephrase it a bit: "Do not keep credentials..." or add an extra panic when it happens?

integration/integration_test.go
mtojek: ~~is it intended to be used anywhere else in envbuilder? I'm considering moving it to `integration/gittest.go`~~

Ok, I found the place. Speaking of consistency with coder/coder that should land in `testutil`.

git_test.go
```diff
@@ -4,72 +4,198 @@ import (
...
+func mustRead(t *testing.T, fs billy.Filesystem, path string) string {
+	t.Helper()
+	f, err := fs.OpenFile(path, os.O_RDONLY, 0644)
+	require.NoError(t, err)
+	content, err := io.ReadAll(f)
+	require.NoError(t, err)
+	return string(content)
+}
+
+func setupGit(t *testing.T, user, pass string) (url string) {
```
mtojek, line 178: What about moving this to `gittest`?

johnstcn: I'll open a a separate PR to refactor this 👍

integration/integration_test.go
johnstcn: Will do in a follow-up PR 👍

git_test.go
```diff
@@ -4,72 +4,198 @@ import (
...
+			password:    "",
+			expectClone: true,
+		},
+		{
+			name:        "password only",
+			srvUsername: "",
+			srvPassword: "password",
+			username:    "",
+			password:    "password",
+			expectClone: true,
```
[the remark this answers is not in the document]
johnstcn: Hm, good point. Once you get into these sort of weeds your testing the tests. Will remove them.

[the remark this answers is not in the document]
johnstcn: We'll keep user-only requests as that's sort of how the GitHub/Azure DevOps authentiation works.

envbuilder.go
johnstcn: I'll just link to the PR with a short note about previous behaviour.

gittest/gittest.go
johnstcn: whoops!

git_test.go
johnstcn: Beefed up the regexp a bit

mtojek (approved):
Thanks for addressing the comments so quickly!

PR #147, opened 2024-04-25
feat: auto add flags docs into the README

BrunoQuaresma:
I still need to make it run on CI and add it to Makefile but I would appreciate an early review of the code. 

You can see a preview of the generated docs here: https://github.com/coder/envbuilder/blob/cd219d378179f0861903190dcd66c117dcc8dd68/README.md#environment-variables

**Strategy**
I added some markers, as HTML comments, in the markdown so the script can know where to place/replace the new content. Another option is to create another markdown file only for the flags and mention it on the main README. 

Close https://github.com/coder/envbuilder/issues/130


README.md
```diff
@@ -243,4 +241,41 @@ On MacOS or Windows systems, we recommend either using a VM or the provided `.de
...
\ No newline at end of file
+- `test-registry`: stands up a local registry for caching images used in tests.
+
+<!--- PLEASE DO NOT MODIFY THIS SECTION. IT IS AUTOMATICALLY GENERATED. --->
+<!--- start-flags --->
+## Flags
+
+| Flag | Env | Default | Description |
+| - | - | - | - |
+| setup-script | SETUP_SCRIPT |  | SetupScript is the script to run before the init script. It runs as the root user regardless of the user specified in the devcontainer.json file. SetupScript is ran as the root user prior to the init script. It is used to configure envbuilder dynamically during the runtime. e.g. specifying whether to start systemd or tiny init for PID 1. |
```
matifali, line 252:
```suggestion
| `--setup-script` | `SETUP_SCRIPT` |  | SetupScript is the script to run before the init script. It runs as the root user regardless of the user specified in the devcontainer.json file. SetupScript is ran as the root user prior to the init script. It is used to configure envbuilder dynamically during the runtime. e.g. specifying whether to start systemd or tiny init for PID 1. |
```
Should we prefix flags with `--` and change style to code?

options.go
```diff
@@ -265,3 +265,15 @@ func (o *Options) CLI() serpent.OptionSet {
...
 	}
 }
+
+func (o *Options) Markdown() string {
+	cliOptions := o.CLI()
+	mkd := "| Flag | Env | Default | Description |\n" +
+		"| - | - | - | - |\n"
+
+	for _, opt := range cliOptions {
+		mkd += "| " + opt.Flag + " | " + opt.Env + " | " + opt.Default + " | " + opt.Description + " |\n"
```
matifali, line 275:
```suggestion
		mkd += "| `--" + opt.Flag + "` | `" + opt.Env + "` | " + opt.Default + " | " + opt.Description + " |\n"
```
I am not sure if this is what we need :D

dannykopping: Came here to say this as well 👍

scripts/docsgen/main.go
```diff
@@ -0,0 +1,34 @@
...
+		panic("Error reading " + readmePath + " file")
+	}
+	readmeContent := string(readmeFile)
+	startIndex := strings.Index(readmeContent, "<!--- start-flags --->")
+	endIndex := strings.Index(readmeContent, "<!--- end-flags --->")
+	if startIndex == -1 || endIndex == -1 {
+		panic("Start or end marks not found in the file.")
+	}
+
+	options := envbuilder.Options{}
```
dannykopping, line 24: This is the Go style for zero-value instantiations.

```suggestion
	var options envbuilder.Options
```

README.md
```diff
@@ -243,4 +241,41 @@ On MacOS or Windows systems, we recommend either using a VM or the provided `.de
 - `build`: builds and tags `envbuilder:latest` for your current architecture.
 - `develop`: runs `envbuilder:latest` against a sample Git repository.
 - `test`: run tests.
-- `test-registry`: stands up a local registry for caching images used in tests.
\ No newline at end of file
+- `test-registry`: stands up a local registry for caching images used in tests.
+
+<!--- PLEASE DO NOT MODIFY THIS SECTION. IT IS AUTOMATICALLY GENERATED. --->
+<!--- start-flags --->
+## Flags
```
dannykopping, line 248:
```suggestion
## CLI Flags
```

dannykopping (approved):

README.md
johnstcn: From reviewing the existing README and codebase, I don't believe that flag-style configuration is supported by `envbuilder`; only environment variables. This makes sense, given that it is primarily meant to be run inside a container.
Adding a `--` prefix will confuse users who will expect this style of invocation to work.

README.md
johnstcn: I would suggest renaming this "Environment Variables" for clarity.

README.md
```diff
@@ -194,8 +194,7 @@ A sample script to pre-fetch a number of images can be viewed [here](./examples/
 
 The `SETUP_SCRIPT` environment variable dynamically configures the user and init command (PID 1) after the container build process.
 
-> **Note**
-> `TARGET_USER` is passed to the setup script to specify who will execute `INIT_COMMAND` (e.g., `code`).
+> **Note** > `TARGET_USER` is passed to the setup script to specify who will execute `INIT_COMMAND` (e.g., `code`).
```
mafredri, line 197:
```suggestion
> [!NOTE]
> `TARGET_USER` is passed to the setup script to specify who will execute `INIT_COMMAND` (e.g., `code`).
```

README.md
```diff
@@ -243,4 +241,41 @@ On MacOS or Windows systems, we recommend either using a VM or the provided `.de
 - `build`: builds and tags `envbuilder:latest` for your current architecture.
 - `develop`: runs `envbuilder:latest` against a sample Git repository.
 - `test`: run tests.
-- `test-registry`: stands up a local registry for caching images used in tests.
\ No newline at end of file
+- `test-registry`: stands up a local registry for caching images used in tests.
+
+<!--- PLEASE DO NOT MODIFY THIS SECTION. IT IS AUTOMATICALLY GENERATED. --->
```
mafredri, line 246:
```suggestion
<!--- Code generated by docsgen. DO NOT EDIT. --->
```
This is a standard format for Go generated files, a bit lossy compared to original but might be nice to retain its property.

README.md
```diff
@@ -243,4 +241,41 @@ On MacOS or Windows systems, we recommend either using a VM or the provided `.de
...
+<!--- start-flags --->
```
mafredri, line 247: I would suggest a more prominent style here, like:
```suggestion
<!--- START docsgen --->
```

And `<!-- END docsgen -->`.

scripts/docsgen/main.go
```diff
@@ -0,0 +1,34 @@
...
+	"strings"
+
+	"github.com/coder/envbuilder"
+)
+
+func main() {
+	readmePath := "README.md"
+	readmeFile, err := os.ReadFile(readmePath)
+	if err != nil {
+		panic("Error reading " + readmePath + " file")
```
mafredri, line 15:
```suggestion
		panic("error reading " + readmePath + " file")
```
Using lower case is more idiomatic, see e.g. coder/coder:

```
~/Work/Code/coder main* ⇣
❯ rg 'panic\([a-z]' | wc -l
      94

~/Work/Code/coder main* ⇣
❯ rg 'panic\([A-Z]' | wc -l
       0
```

scripts/docsgen/main.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+	readmeContent := string(readmeFile)
+	startIndex := strings.Index(readmeContent, "<!--- start-flags --->")
```
mafredri, line 18: Perhaps these could be defined at the top of the file as:

```
const (
	startSection = "<!-- ..."
	endSection = "<!-- ..."
)
```

Easier maintenance if changing it in the future.

README.md
BrunoQuaresma: @johnstcn `coder/serpent` supports flags out of the box like env variables so in theory, it should work.

scripts/docsgen/main.go
BrunoQuaresma: I'm wondering if it has any specific reason for lowercase to be considered more idiomatic.

BrunoQuaresma:
Updated preview: https://github.com/coder/envbuilder/blob/cd219d378179f0861903190dcd66c117dcc8dd68/README.md

README.md
```diff
@@ -243,4 +242,41 @@ On MacOS or Windows systems, we recommend either using a VM or the provided `.de
...
+- `test-registry`: stands up a local registry for caching images used in tests.
+
+<!--- Code generated by docsgen. DO NOT EDIT. --->
+<!--- START docsgen --->
+## Environment Variables
+
+| Environment variable | Default | Description |
+| - | - | - |
+| SETUP_SCRIPT |  | SetupScript is the script to run before the init script. It runs as the root user regardless of the user specified in the devcontainer.json file. SetupScript is ran as the root user prior to the init script. It is used to configure envbuilder dynamically during the runtime. e.g. specifying whether to start systemd or tiny init for PID 1. |
+| INIT_SCRIPT | sleep infinity | InitScript is the script to run to initialize the workspace. |
```
mafredri, line 254: I think it would look nice if these were inline code blocks, like this:
```suggestion
| `INIT_SCRIPT` | `sleep infinity` | InitScript is the script to run to initialize the workspace. |
```
(For "Default", we'd only insert `` if it's a non-empty value.)

mafredri (approved):
LGTM!

README.md
BrunoQuaresma: Done!

matifali (approved):

Issue #144, opened 2024-04-25
Running envbuilder CLI on a local laptop may destroy the filesystem


bpmct:
Can we add a warning that explains the CLI is not intended to be ran locally on a laptop (Macbook) and a warning of the outcome?

VM use cases are encouraged, but we should still warn

bpmct:
FYI @mtojek

matifali:
I have been through this :( Lost all my data.

mtojek:
> I have been through this :( Lost all my data.

Can you share more details about what exactly happened? We may describe this in README.

matifali:
I was logged in through SSH to a Linux VM and executed the `envbuilder` binary. I got disconnected and could not log in again. I assume it destroyed/overridden the filesystem.

johnstcn:
While VM use-cases for envbuilder are definitely possible, its primary use-case is to run inside a container.

johnstcn:
> I was logged in through SSH to a Linux VM and executed the `envbuilder` binary. I got disconnected and could not log in again. I assume it destroyed/overrode the filesystem.

I was able to reproduce this, but needed to run `envbuilder` as `root`.

johnstcn:
Root cause: `util.DeleteFilesystem` will delete `KANIKO_DIR`.
In the envbuilder Dockerfile, we set `KANIKO_DIR=/.envbuilder`.
Otherwise, `KANIKO_DIR` defaults to `/`.

PR #153, opened 2024-04-26
feat!: add safety check if not running as PID 1.

johnstcn:
Relates to https://github.com/coder/envbuilder/issues/144
This PR adds a safety check to guard against users accidentally running envbuilder on their local system.
It can be bypassed by setting an environment variable.
This is a breaking change for use-cases outside of containers.

### DIFF
--- a/cmd/envbuilder/main.go
+++ b/cmd/envbuilder/main.go
@@ -33,12 +33,12 @@ func main() {
 			options := envbuilder.OptionsFromEnv(os.LookupEnv)
 			if os.Getpid() != 1 {
 				// TODO: rebase once https://github.com/coder/envbuilder/pull/140 is in
-				if env, found := os.LookupEnv("DANGEROUS_BYPASS_PID_CHECK"); found && env == "1" {
-					_, _ = fmt.Fprintln(os.Stderr, `Bypassing PID check as DANGEROUS_BYPASS_PID_CHECK=1.`)
+				if env, found := os.LookupEnv("DANGEROUS_DISABLE_PID_CHECK"); found && env == "1" {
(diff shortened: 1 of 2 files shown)


mtojek (approved):
> Relates to

I guess it resolves it?

cmd/envbuilder/main.go
```diff
@@ -31,6 +31,17 @@ func main() {
 		SilenceErrors: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			options := envbuilder.OptionsFromEnv(os.LookupEnv)
+			if os.Getpid() != 1 {
+				// TODO: rebase once https://github.com/coder/envbuilder/pull/140 is in
+				if env, found := os.LookupEnv("DANGEROUS_DISABLE_PID_CHECK"); found && env == "1" {
+					_, _ = fmt.Fprintln(os.Stderr, `Bypassing PID check as DANGEROUS_DISABLE_PID_CHECK=1.`)
+				} else {
```
johnstcn, line 38: review: will move this inside `Options` once Bruno's PR is in.

mtojek:
> There's also other ways to detect we're in a container and (allow?) execution even if not PID 1.

It is an interesting point. Do we intend to support running envbuilder in a different way than using the official container image?

johnstcn:
> > There's also other ways to detect we're in a container and (allow?) execution even if not PID 1.
> 
> It is an interesting point. Do we intend to support running envbuilder in a different way than using the official container image?

At present, we only distribute envbuilder via Docker image. We do not provide precompiled binaries outside of this.

johnstcn:
Looking further at the options, `FORCE_SAFE` should be handling this.

johnstcn:
Opening a separate PR with the actual correct fix.

PR #160, opened 2024-04-29
fix: avoid deleting root filesystem when KANIKO_DIR not set

johnstcn:
Fixes https://github.com/coder/envbuilder/issues/144

Running `envbuilder` outside of a container as root can result in it nuking your entire FS if you're not careful.

Adds a check if `KANIKO_DIR` is not set to `MagicDir` (a.k.a. `/.envbuilder`) and bails early.
This behaviour can be overridden with `FORCE_SAFE=true`.


README.md
```diff
@@ -47,6 +47,11 @@ $ vim .devcontainer/Dockerfile
 
 Exit the container, and re-run the `docker run` command... after the build completes, `htop` should exist in the container! 🥳
 
+> **Note:** Envbuilder performs destructive filesystem operations! To guard against accidental data
```
mafredri, line 50:
```suggestion
> [!NOTE]
> Envbuilder performs destructive filesystem operations! To guard against accidental data
```

README.md
```diff
@@ -47,6 +47,11 @@ $ vim .devcontainer/Dockerfile
...
+> loss, it will refuse to run if it detects that KANIKO_DIR is not set to a specific value.
+> If you need to bypass this behaviour for any reason, you can bypass this safety check by setting
```
mafredri, line 52:
```suggestion
> If you need to bypass this behavior for any reason, you can bypass this safety check by setting
```
😏

README.md
```diff
@@ -47,6 +47,11 @@ $ vim .devcontainer/Dockerfile
...
+> `FORCE_SAFE=true`.
```
mafredri, line 53: I'd love for us to start using the `ENVBUILDER_` prefix for our envs, can we start now?

envbuilder.go
```diff
@@ -1063,3 +1062,20 @@ func findDevcontainerJSON(options Options) (string, string, error) {
 
 	return "", "", errors.New("can't find devcontainer.json, is it a correct spec?")
 }
+
+// maybeDeleteFilesystem wraps util.DeleteFilesystem with a guard to hopefully stop
+// folks from unwittingly deleting their entire root directory.
+func maybeDeleteFilesystem(force bool) error {
```
mafredri, line 1068: I love this name ❤️

dannykopping (approved):
LGTM

envbuilder.go
```diff
@@ -1063,3 +1062,20 @@ func findDevcontainerJSON(options Options) (string, string, error) {
...
+	kanikoDir, ok := os.LookupEnv("KANIKO_DIR")
+	if !ok || strings.TrimSpace(kanikoDir) != MagicDir {
+		if force {
+			_, _ = fmt.Fprintf(os.Stderr, "WARNING! BYPASSING SAFETY CHECK! THIS WILL DELETE %s!\n", kanikoDir)
```
mafredri, line 1072: Idea: how about `time.Sleep(2*time.Second)`, "you have to seconds to abort"?

mafredri (approved):

envbuilder.go
johnstcn: lol I was actually thinking about doing this. I'd go for ten seconds just to be sure.

README.md
johnstcn: I'll do this in a separate PR. This isn't a new option, it's been there all the time but not doing anything.

Issue #157, opened 2024-04-29
Test to validate that envs given previously will still produce the same options now in Serpent


BrunoQuaresma:
Related to: https://github.com/coder/envbuilder/pull/140
Reference: https://github.com/coder/envbuilder/pull/140#discussion_r1581024599

PR #161, opened 2024-04-29
test: add test for options env

BrunoQuaresma:
After some thinking, I concluded that we don't need to test all the env vars but how those values are parsed, but let me know if you think we should test all the env variables.

Closes https://github.com/coder/envbuilder/issues/157

### DIFF
--- a/options_test.go
+++ b/options_test.go
@@ -12,15 +12,14 @@ import (
 // TestEnvOptionParsing tests that given environment variables of different types are handled as expected.
 func TestTestEnvOptionParsing(t *testing.T) {
 	t.Run("string", func(t *testing.T) {
-		t.Setenv("SETUP_SCRIPT", "setup.sh")
-		t.Setenv("INIT_COMMAND", "sleep infinity")
+	        const val = "setup.sh"
+		t.Setenv("SETUP_SCRIPT", val)
 
 		var o envbuilder.Options
 		err := runCLI(&o)
 		require.NoError(t, err)
 
-		require.Equal(t, o.SetupScript, "setup.sh")
-		require.Equal(t, o.InitCommand, "sleep infinity")
+		require.Equal(t, o.SetupScript, val)
 	})
 
 	t.Run("int", func(t *testing.T) {
--- a/options_test.go
+++ b/options_test.go
@@ -10,68 +10,77 @@ import (
 )
 
 // TestEnvOptionParsing tests that given environment variables of different types are handled as expected.
-func TestTestEnvOptionParsing(t *testing.T) {
+func TestEnvOptionParsing(t *testing.T) {
 	t.Run("string", func(t *testing.T) {
-	        const val = "setup.sh"
+		const val = "setup.sh"
 		t.Setenv("SETUP_SCRIPT", val)
-
-		var o envbuilder.Options
-		err := runCLI(&o)
-		require.NoError(t, err)
-
+		o := runCLI()
 		require.Equal(t, o.SetupScript, val)
 	})
 
 	t.Run("int", func(t *testing.T) {
 		t.Setenv("CACHE_TTL_DAYS", "7")
-
-		var o envbuilder.Options
-		err := runCLI(&o)
-		require.NoError(t, err)
-
+		o := runCLI()
 		require.Equal(t, o.CacheTTLDays, int64(7))
 	})
 
 	t.Run("string array", func(t *testing.T) {
 		t.Setenv("IGNORE_PATHS", "/var,/temp")
-
-		var o envbuilder.Options
-		err := runCLI(&o)
-		require.NoError(t, err)
-
+		o := runCLI()
 		require.Equal(t, o.IgnorePaths, []string{"/var", "/temp"})
 	})
 
 	t.Run("bool", func(t *testing.T) {
-		t.Setenv("SKIP_REBUILD", "true")
-		t.Setenv("GIT_CLONE_SINGLE_BRANCH", "false")
-		t.Setenv("EXIT_ON_BUILD_FAILURE", "true")
-		t.Setenv("FORCE_SAFE", "false")
-		t.Setenv("INSECURE", "true")
+		t.Run("lowercase", func(t *testing.T) {
+			t.Setenv("SKIP_REBUILD", "true")
+			t.Setenv("GIT_CLONE_SINGLE_BRANCH", "false")
+			o := runCLI()
+			require.True(t, o.SkipRebuild)
+			require.False(t, o.GitCloneSingleBranch)
+		})
 
-		var o envbuilder.Options
-		err := runCLI(&o)
-		require.NoError(t, err)
+		t.Run("uppercase", func(t *testing.T) {
+			t.Setenv("SKIP_REBUILD", "TRUE")
+			t.Setenv("GIT_CLONE_SINGLE_BRANCH", "FALSE")
+			o := runCLI()
+			require.True(t, o.SkipRebuild)
+			require.False(t, o.GitCloneSingleBranch)
+		})
 
-		require.True(t, o.SkipRebuild)
-		require.False(t, o.GitCloneSingleBranch)
-		require.True(t, o.ExitOnBuildFailure)
-		require.False(t, o.ForceSafe)
-		require.True(t, o.Insecure)
+		t.Run("numeric", func(t *testing.T) {
+			t.Setenv("SKIP_REBUILD", "1")
+			t.Setenv("GIT_CLONE_SINGLE_BRANCH", "0")
+			o := runCLI()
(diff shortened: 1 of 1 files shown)


options_test.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func TestEnvOptions(t *testing.T) {
+	t.Setenv("SETUP_SCRIPT", "setup.sh")
+	t.Setenv("CACHE_TTL_DAYS", "7")
+	t.Setenv("IGNORE_PATHS", "/var,/tmp")
+	t.Setenv("SKIP_REBUILD", "true")
+	t.Setenv("GIT_CLONE_SINGLE_BRANCH", "")
+
+	var o = envbuilder.Options{}
```
dannykopping, line 25: Nit
```suggestion
	var o envbuilder.Options
```

It's currently somewhere between `o := envbuilder.Options{}` and the suggestion above
We should pick one, I don't mind which

options_test.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+type testEnvCase struct {
+	Env      string
+	EnvValue string
+	Expected interface{}
+	Actual   func() interface{}
+}
+
+func TestEnvOptions(t *testing.T) {
```
dannykopping, line 18:
```suggestion
// TestEnvOptionParsing tests that given environment variables of different types are handled as expected.
func TestEnvOptionParsing(t *testing.T) {
```

options_test.go
```diff
@@ -0,0 +1,40 @@
...
+	t.Setenv("SETUP_SCRIPT", "setup.sh")
+	t.Setenv("CACHE_TTL_DAYS", "7")
+	t.Setenv("IGNORE_PATHS", "/var,/tmp")
+	t.Setenv("SKIP_REBUILD", "true")
```
dannykopping, line 22: The previous code was using https://pkg.go.dev/strconv#ParseBool, but I'm not sure what the new code is using so we might need to test `1`/`0`/`true`/`false` (and the latter two's capitalised equivalents).

options_test.go
```diff
@@ -0,0 +1,40 @@
...
+		Options: o.CLI(),
+		Handler: func(inv *serpent.Invocation) error {
+			return nil
+		},
+	}
+	err := cmd.Invoke().WithOS().Run()
+	require.NoError(t, err)
+
+	require.Equal(t, o.SetupScript, "setup.sh")
+	require.Equal(t, o.CacheTTLDays, int64(7))
```
dannykopping, line 36: I'd suggest choosing a large number to validate that the current code is handling integers as 64-bit

options_test.go
```diff
@@ -0,0 +1,40 @@
...
+	require.Equal(t, o.IgnorePaths, []string{"/var", "/tmp"})
```
dannykopping, line 37: Were only string slices accepted?

options_test.go
BrunoQuaresma: Hmm.. we currently use just lower numbers, to be honest. In this case, using `int64` is a `coder/serpent` thing.

Emyrk:
The `runCLI` is doing a lot of extra work, where this is really just testing env var assignment. An option instead is to skip the cli part, and just do the env parsing:

```golang
	t.Run("string", func(t *testing.T) {
		var o envbuilder.Options
		set := o.CLI()
		err := set.ParseEnv([]serpent.EnvVar{
			{"SETUP_SCRIPT", "setup.sh"},
			{"INIT_COMMAND", "sleep infinity"},
		})
		require.NoError(t, err)

		require.Equal(t, o.SetupScript, "setup.sh")
		require.Equal(t, o.InitCommand, "sleep infinity")
	})
```

If we do this, we can also just merge all the tests. But one thing that feels off here is that we are only testing the happy paths, which makes this test feel weak in general.

We could comment this is just a sanity check to make sure we don't change env vars without documenting it to prevent any breaking changes. In that case, I think merging the tests and doing it the method above would be a bit cleaner.


```golang
// TestEnvOptionParsing prevents breaking changes to the environment variables.
func TestEnvOptionParsing(t *testing.T) {
	t.Parallel()

	var o envbuilder.Options
	set := o.CLI()
	err := set.ParseEnv([]serpent.EnvVar{
		{"SETUP_SCRIPT", "setup.sh"},
		{"INIT_COMMAND", "sleep infinity"},
		// ... Add the rest
		{"CACHE_TTL_DAYS", "7"},
	})
	require.NoError(t, err)

	require.Equal(t, o.SetupScript, "setup.sh")
	require.Equal(t, o.InitCommand, "sleep infinity")
 	// ... Add the rest
	require.Equal(t, o.CacheTTLDays, int64(7))
}
```

BrunoQuaresma:
@Emyrk I think it makes sense. Since it is a straightforward test I don't see how to test "unhappy" paths, how would you do that in this scenario?

options_test.go
```diff
@@ -1,40 +1,90 @@
...
+		t.Setenv("SETUP_SCRIPT", "setup.sh")
+		t.Setenv("INIT_COMMAND", "sleep infinity")
+
+		var o envbuilder.Options
+		err := runCLI(&o)
+		require.NoError(t, err)
+
+		require.Equal(t, o.SetupScript, "setup.sh")
+		require.Equal(t, o.InitCommand, "sleep infinity")
+	})
```
dannykopping, line 24: I think this expresses the intent more concisely, but to be clear your test as it stands is perfectly fine as it is.

```suggestion
	t.Run("string", func(t *testing.T) {
	        const val = "setup.sh"
		t.Setenv("SETUP_SCRIPT", val)

		var o envbuilder.Options
		err := runCLI(&o)
		require.NoError(t, err)

		require.Equal(t, o.SetupScript, val)
	})
```

options_test.go
```diff
@@ -1,40 +1,90 @@
...
+		require.Equal(t, o.IgnorePaths, []string{"/var", "/temp"})
+	})
 
-	var o = envbuilder.Options{}
+	t.Run("bool", func(t *testing.T) {
+		t.Setenv("SKIP_REBUILD", "true")
+		t.Setenv("GIT_CLONE_SINGLE_BRANCH", "false")
+		t.Setenv("EXIT_ON_BUILD_FAILURE", "true")
+		t.Setenv("FORCE_SAFE", "false")
+		t.Setenv("INSECURE", "true")
```
dannykopping, line 51: You're currently only testing one sort of coercion: lowecase string->bool. There are others which are supported by https://pkg.go.dev/strconv#ParseBool.

```suggestion
		t.Setenv("SKIP_REBUILD", "TRUE")
		t.Setenv("GIT_CLONE_SINGLE_BRANCH", "FALSE")
		t.Setenv("EXIT_ON_BUILD_FAILURE", "1")
		t.Setenv("FORCE_SAFE", "0")
		t.Setenv("INSECURE", "true")
```

options_test.go
dannykopping: Looking at the current options for envbuilder, it seems like only slices of strings are accepted so this is fine 👍

dannykopping (commented):
Almost there, thanks @BrunoQuaresma 👍

johnstcn (approved):

options_test.go
BrunoQuaresma: What if we don't have enough options to test all of them? 😅

options_test.go
BrunoQuaresma: I like this way but when having to test 6 options, it looks too much IMO. Wdyt?

options_test.go
dannykopping: Looks like we have enough to test all those cases?

options_test.go
BrunoQuaresma: It would looks like:

```go
const skipRebuildValue = true
const gitCloneSingleBranchValue = false
const exitOnBuildFailureValue = false
// and so on...

```

options_test.go
BrunoQuaresma: it would be missing:
- Empty value as false
- "false" in lowercase as false

options_test.go
dannykopping: Oh no I just meant for this particular test; don't worry about applying it to the others.

options_test.go
dannykopping: Ah. You could test the same options below this set of tests, or break them up into sections?

Like:

`true`<br>
`false`

---

`TRUE`<br>
`FALSE`

---

`1`
`0`

---

`<empty>`

---

With things like this it's better to be comprehensive rather than a customer running into strange behaviour which is difficult to track down

Emyrk (commented):
This looks fine to me. I think we should port a more rigorous test like this to the serpent package.

Side note, do we not have a linter on this project?

dannykopping:
Good work @BrunoQuaresma! Thanks for being patient with the review.

Issue #159, opened 2024-04-29
Warn if clean up does not work when removing Docker config file.


BrunoQuaresma:
Related to: https://github.com/coder/envbuilder/pull/140
Reference: https://github.com/coder/envbuilder/pull/140#discussion_r1581217631

PR #163, opened 2024-04-30
chore: warn if docker secret cleanup fails

BrunoQuaresma:
Closes https://github.com/coder/envbuilder/issues/159

### DIFF
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -604,9 +604,14 @@ func Run(ctx context.Context, options Options) error {
 
 	// Remove the Docker config secret file!
 	if options.DockerConfigBase64 != "" {
-		err = os.Remove(filepath.Join(MagicDir, "config.json"))
-		if err != nil && !os.IsNotExist(err) {
-			return fmt.Errorf("remove docker config: %w", err)
+		c := filepath.Join(MagicDir, "config.json")
+		err = os.Remove(c)
+		if err != nil {
+			if !os.IsNotExist(err) {
+				return fmt.Errorf("remove docker config: %w", err)
+			} else {
+				fmt.Fprintln(os.Stderr, "failed to remove the Docker config secret file: %w", c)
+			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


envbuilder.go
```diff
@@ -604,9 +604,14 @@ func Run(ctx context.Context, options Options) error {
...
-		err = os.Remove(filepath.Join(MagicDir, "config.json"))
-		if err != nil && !os.IsNotExist(err) {
-			return fmt.Errorf("remove docker config: %w", err)
+		c := filepath.Join(MagicDir, "config.json")
+		err = os.Remove(c)
+		if err != nil {
+			if !os.IsNotExist(err) {
+				return fmt.Errorf("remove docker config: %w", err)
+			} else {
+				fmt.Fprintln(os.Stderr, "failed to remove the Docker config secret file: %w", c)
```
johnstcn, line 613: Why not use `options.Logger` here and warn?

BrunoQuaresma: I saw both options being used in the code so I didn't know what to use, to be honest.

envbuilder.go
```diff
@@ -604,9 +604,14 @@ func Run(ctx context.Context, options Options) error {
 
 	// Remove the Docker config secret file!
 	if options.DockerConfigBase64 != "" {
-		err = os.Remove(filepath.Join(MagicDir, "config.json"))
-		if err != nil && !os.IsNotExist(err) {
-			return fmt.Errorf("remove docker config: %w", err)
+		c := filepath.Join(MagicDir, "config.json")
+		err = os.Remove(c)
```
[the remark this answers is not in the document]
johnstcn: Let's keep it out of scope of this PR then.

envbuilder.go
johnstcn: That's a fair point. I'll raise a separate PR to standardize this.

johnstcn:
I think this might also be a good place to add a check in the integration tests that the docker config.json was removed, something like this might work:

```
		statOutput := execContainer(t, ctr, "stat "+filepath.Join(envbuilder.MagicDir, "config.json"))
		require.Contains(t, statOutput, "No such file or directory")
```

johnstcn (approved):

Issue #31, opened 2023-07-14
SSH repositories fail to configure with coder


geiseri:
If I try to use a private git repository over SSH I see the following error:

```
envbuilder - Build development environments from repositories in a container
#1: 📦 Cloning ssh://git@github.com/hive-io/hive-rest.git to /workspaces/hive-rest.git...
Failed to clone repository: clone "ssh://git@github.com/hive-io/hive-rest.git": error creating SSH agent: "SSH agent requested but SSH_AUTH_SOCK not-specified"
Falling back to the default image...
#2: 🏗️ Building image...
#2: Retrieving image manifest codercom/enterprise-base:ubuntu
#2: Retrieving image codercom/enterprise-base:ubuntu from registry index.docker.io
#2: Built cross stage deps: map[]
```

I didn't see a way to configure that to use the built in SSH key provided by coder.  Is there a missing setting?

janLo:
Would it be sufficient to drop an coder agent binary into the envbuilder image and set `GIT_SSH_COMMAND=/.envbuilder/coder gitssh --` ?

chrisspalm:
This would be useful!

geiseri:
Those are in the user profile right?

chewbh:
Is there plan to look at implementing this? Support for cloning over SSH in envbuilder will be very useful. 

Some developers has preference using SSH over HTTP/S for interacting with git. To them, it is not expected that a devcontainer workspace fail to build due to this, given that they have already configure a SSH key within Coder and that they are force to fallback to use HTTP/S for cloning. 

In addition, to support cloning private repos, we specifically need to add additional terraform logic or uses coder_external_auth  to configure GIT environment variables in the workspace template.

nwrkbiz:
Adding support for this, would also provide a simple and secure workaround for #60

thattolleyguy:
+1 for adding support for this. This is the #1 reason our company isn't using coder yet

johnstcn:
Plan:
- Add support for reading a SSH private key from a file (SSH_KEY_PATH)
- Add support for passing in a known_hosts file (SSH_KNOWN_HOSTS_BASE64)
- Add support for performing a SSH keyscan and generating a known_hosts file if no known_hosts file provided
- Add support for SSH key authentication using SSH_KEY_PATH
[...]

PR #170, opened 2024-05-02
feat: support cloning over SSH via private key auth

johnstcn:
Relates to https://github.com/coder/envbuilder/issues/31

Currently running into issues with testing end-to-end flow.
However, I've tested with a private git repo:

```
$ docker run -it --rm -e GIT_URL="ssh://git@github.com/johnstcn/test-private-devcontainer" -e INIT_SCRIPT=bash -e GIT_SSH_PRIVATE_KEY_PATH=/.ssh/id_ed25519 -v /Users/cian/.ssh/id_ed25519:/.ssh/id_ed25519 envbuilder:latest
envbuilder - Build development environments from repositories in a container
ssh keyscan: github.com ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBEmKSENjQEezOmxkZMy7opKgwFB9nkt5YRrYMjNuG5N87uRgg6CLrbo5wAdT/y6v0mKV0U2w0WZ2YB/++Tpockg=
#1: 📦 Cloning ssh://git@github.com/johnstcn/test-private-devcontainer to /workspaces/test-private-devcontainer...
#1: Enumerating objects: 6, done.
#1: Counting objects:  16% (1/6)
#1: Counting objects:  33% (2/6)
#1: Counting objects:  50% (3/6)
#1: Counting objects:  66% (4/6)
#1: Counting obj
#1: ects:  83% (5/6)
#1: Counting objects: 100% (6/6)
#1: Counting objects: 100% (6/6), done.
#1: Compressing objects:  16% (1/6)
#1: Compressing objects:  33% (2/6)
#1: Compressing objects:  50% (3/6)
#1: Compressing objects:  66% (4/6)
#1: Compressing objects:  83% (5/6)
#1: Compressing objects: 100% (6/6)
#1: Compressing objects: 100% (6/6), done.
#1: Total 6 (delta 0), reused 6 (delta 0), pack-reused 0
#1: 📦 Cloned repository! [1.434724587s]
```

One potential issue for which I don't have a solution is that passing an SSH key in as a volume will result in the file being available in the resulting workspace. This may not be desirable; an alternative would be to pass in the SSH key as a base64-encoded file (similar to known_hosts) and clean it up after the build completes.

After this is merged, I'd like to do a follow-up PR to refactor some of the envbuilder setup code.

### DIFF
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -44,8 +44,6 @@ import (
 	"github.com/go-git/go-billy/v5"
 	"github.com/go-git/go-billy/v5/osfs"
 	"github.com/go-git/go-git/v5/plumbing/transport"
-	githttp "github.com/go-git/go-git/v5/plumbing/transport/http"
-	gitssh "github.com/go-git/go-git/v5/plumbing/transport/ssh"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/sirupsen/logrus"
@@ -157,36 +155,6 @@ func Run(ctx context.Context, options Options) error {
 		}
 	}
 
-	if options.GitURL != "" {
-		gitURLParsed, err := ParseGitURL(options.GitURL)
-		if err != nil {
-			return fmt.Errorf("invalid git URL: %w", err)
-		}
-		// If we're cloning over SSH, we need a known_hosts file.
-		if gitURLParsed.Scheme == "ssh" {
-			var knownHostsContent []byte
-			if options.GitSSHKnownHostsBase64 != "" {
-				if kh, err := base64.StdEncoding.DecodeString(options.GitSSHKnownHostsBase64); err != nil {
-					return fmt.Errorf("invalid known_hosts content: %w", err)
-				} else {
-					knownHostsContent = kh
-				}
-			} else {
-				// This is a best-effort.
-				kh, err := KeyScan(options.Logger, gitURLParsed)
-				if err == nil {
-					knownHostsContent = kh
-				}
-			}
-			knownHostsPath := filepath.Join(MagicDir, "known_hosts")
-			if err := os.WriteFile(knownHostsPath, knownHostsContent, 0644); err != nil {
-				return fmt.Errorf("write known_hosts file: %w", err)
-			}
-			// go-git will read this file to validate the server host keys.
-			_ = os.Setenv("SSH_KNOWN_HOSTS", knownHostsPath)
-		}
-	}
-
 	var fallbackErr error
 	var cloned bool
 	if options.GitURL != "" {
@@ -225,23 +193,7 @@ func Run(ctx context.Context, options Options) error {
 			CABundle:     caBundle,
 		}
 
-		if options.GitUsername != "" || options.GitPassword != "" {
-			// NOTE: we previously inserted the credentials into the repo URL.
-			// This was removed in https://github.com/coder/envbuilder/pull/141
-			cloneOpts.RepoAuth = &githttp.BasicAuth{
-				Username: options.GitUsername,
-				Password: options.GitPassword,
-			}
-		} else if options.GitSSHPrivateKeyPath != "" {
-			signer, err := ReadPrivateKey(options.GitSSHPrivateKeyPath)
-			if err != nil {
-				return xerrors.Errorf("read private key: %w", err)
-			}
-			cloneOpts.RepoAuth = &gitssh.PublicKeys{
-				User:   "git",
-				Signer: signer,
-			}
-		}
+		cloneOpts.RepoAuth = SetupRepoAuth(&options)
 		if options.GitHTTPProxyURL != "" {
 			cloneOpts.ProxyOptions = transport.ProxyOptions{
 				URL: options.GitHTTPProxyURL,
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -171,7 +171,7 @@ func Run(ctx context.Context, options Options) error {
 				knownHostsContent = kh
 			}
 		} else {
-			kh, err := GenerateKnownHosts(options.Logger, gitURLParsed)
+			kh, err := KeyScan(options.Logger, gitURLParsed)
 			if err != nil {
 				return fmt.Errorf("invalid known_hosts content: %w", err)
 			} else {
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -157,7 +157,7 @@ func Run(ctx context.Context, options Options) error {
 		}
 	}
 
-	gitURLParsed, err := url.Parse(options.GitURL)
+	gitURLParsed, err := ParseGitURL(options.GitURL)
 	if err != nil {
 		return fmt.Errorf("invalid git URL: %w", err)
 	}
@@ -182,6 +182,7 @@ func Run(ctx context.Context, options Options) error {
 		if err := os.WriteFile(knownHostsPath, knownHostsContent, 0644); err != nil {
 			return fmt.Errorf("write known_hosts file: %w", err)
 		}
+		// go-git will read this file to validate the server host keys.
 		_ = os.Setenv("SSH_KNOWN_HOSTS", knownHostsPath)
 	}
 
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -45,6 +45,7 @@ import (
 	"github.com/go-git/go-billy/v5/osfs"
 	"github.com/go-git/go-git/v5/plumbing/transport"
 	githttp "github.com/go-git/go-git/v5/plumbing/transport/http"
+	gitssh "github.com/go-git/go-git/v5/plumbing/transport/ssh"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/sirupsen/logrus"
@@ -156,6 +157,34 @@ func Run(ctx context.Context, options Options) error {
 		}
 	}
 
+	gitURLParsed, err := url.Parse(options.GitURL)
+	if err != nil {
+		return fmt.Errorf("invalid git URL: %w", err)
+	}
+	// If we're cloning over SSH, we need a known_hosts file.
+	if gitURLParsed.Scheme == "ssh" {
+		var knownHostsContent []byte
+		if options.GitSSHKnownHostsBase64 != "" {
+			if kh, err := base64.StdEncoding.DecodeString(options.GitSSHKnownHostsBase64); err != nil {
+				return fmt.Errorf("invalid known_hosts content: %w", err)
+			} else {
+				knownHostsContent = kh
+			}
+		} else {
+			kh, err := GenerateKnownHosts(options.Logger, gitURLParsed)
+			if err != nil {
+				return fmt.Errorf("invalid known_hosts content: %w", err)
+			} else {
+				knownHostsContent = kh
+			}
+		}
+		knownHostsPath := filepath.Join(MagicDir, "known_hosts")
+		if err := os.WriteFile(knownHostsPath, knownHostsContent, 0644); err != nil {
+			return fmt.Errorf("write known_hosts file: %w", err)
+		}
+		_ = os.Setenv("SSH_KNOWN_HOSTS", knownHostsPath)
+	}
+
 	var fallbackErr error
 	var cloned bool
 	if options.GitURL != "" {
@@ -200,6 +229,15 @@ func Run(ctx context.Context, options Options) error {
 			cloneOpts.RepoAuth = &githttp.BasicAuth{
 				Username: options.GitUsername,
 				Password: options.GitPassword,
+			}
+		} else if options.GitSSHPrivateKeyPath != "" {
+			signer, err := ReadPrivateKey(options.GitSSHPrivateKeyPath)
+			if err != nil {
+				return xerrors.Errorf("read private key: %w", err)
+			}
+			cloneOpts.RepoAuth = &gitssh.PublicKeys{
+				User:   "git",
+				Signer: signer,
 			}
 		}
 		if options.GitHTTPProxyURL != "" {
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -157,33 +157,34 @@ func Run(ctx context.Context, options Options) error {
 		}
 	}
 
-	gitURLParsed, err := ParseGitURL(options.GitURL)
-	if err != nil {
-		return fmt.Errorf("invalid git URL: %w", err)
-	}
-	// If we're cloning over SSH, we need a known_hosts file.
-	if gitURLParsed.Scheme == "ssh" {
-		var knownHostsContent []byte
-		if options.GitSSHKnownHostsBase64 != "" {
-			if kh, err := base64.StdEncoding.DecodeString(options.GitSSHKnownHostsBase64); err != nil {
-				return fmt.Errorf("invalid known_hosts content: %w", err)
-			} else {
-				knownHostsContent = kh
-			}
-		} else {
-			kh, err := KeyScan(options.Logger, gitURLParsed)
-			if err != nil {
-				return fmt.Errorf("invalid known_hosts content: %w", err)
-			} else {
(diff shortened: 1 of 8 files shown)


envbuilder.go
```diff
@@ -156,6 +157,34 @@ func Run(ctx context.Context, options Options) error {
...
+				return fmt.Errorf("invalid known_hosts content: %w", err)
+			} else {
+				knownHostsContent = kh
+			}
+		}
+		knownHostsPath := filepath.Join(MagicDir, "known_hosts")
+		if err := os.WriteFile(knownHostsPath, knownHostsContent, 0644); err != nil {
+			return fmt.Errorf("write known_hosts file: %w", err)
+		}
+		_ = os.Setenv("SSH_KNOWN_HOSTS", knownHostsPath)
```
mafredri, line 185: Can we add a comment that this is for use by `go-git`?

envbuilder.go
```diff
@@ -156,6 +157,34 @@ func Run(ctx context.Context, options Options) error {
 		}
 	}
 
+	gitURLParsed, err := url.Parse(options.GitURL)
+	if err != nil {
+		return fmt.Errorf("invalid git URL: %w", err)
+	}
+	// If we're cloning over SSH, we need a known_hosts file.
+	if gitURLParsed.Scheme == "ssh" {
```
mafredri, line 165: I believe it's possible to clone over SSH even without `ssh://`. Just `git@github.com:org/repo` should suffice. I wonder if this will catch that case?

envbuilder.go
```diff
@@ -156,6 +157,34 @@ func Run(ctx context.Context, options Options) error {
...
+		} else {
+			kh, err := GenerateKnownHosts(options.Logger, gitURLParsed)
+			if err != nil {
+				return fmt.Errorf("invalid known_hosts content: %w", err)
+			} else {
+				knownHostsContent = kh
+			}
+		}
+		knownHostsPath := filepath.Join(MagicDir, "known_hosts")
+		if err := os.WriteFile(knownHostsPath, knownHostsContent, 0644); err != nil {
```
mafredri, line 182: Why do we need a file if we set the env?

git.go
```diff
@@ -113,3 +122,55 @@ func CloneRepo(ctx context.Context, opts CloneRepoOptions) (bool, error) {
...
 	return true, nil
 }
+
+func ReadPrivateKey(path string) (gossh.Signer, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, fmt.Errorf("open private key file: %w", err)
+	}
+	var buf bytes.Buffer
+	if _, err := io.Copy(&buf, f); err != nil {
```
mafredri, line 132: Suggestion: `io.ReadAll` would be simpler here.

git.go
```diff
@@ -113,3 +122,55 @@ func CloneRepo(ctx context.Context, opts CloneRepoOptions) (bool, error) {
...
+		},
+	}
+	dialAddr := hostPort(gitURL)
+	client, err := gossh.Dial("tcp", dialAddr, conf)
+	if err != nil {
+		// The dial may fail due to no authentication methods, but this is fine.
+		if netErr, ok := err.(net.Error); ok {
+			return nil, fmt.Errorf("keyscan %s: %w", dialAddr, netErr)
+		}
+		// If it's not a net.Error then we will assume we were successful.
```
mafredri, line 163: Should we perhaps check if buf is empty?

git.go
```diff
@@ -113,3 +122,55 @@ func CloneRepo(ctx context.Context, opts CloneRepoOptions) (bool, error) {
...
+	} else {
+		_ = client.Close()
+	}
+	return buf.Bytes(), nil
+}
+
+func hostPort(u *url.URL) string {
+	p := 22 // assume default SSH port
+	if _p, err := strconv.Atoi(u.Port()); err == nil {
```
mafredri, line 172: Nit: Prefer `pp`, underscores are unusual in variable names.

git_test.go
```diff
@@ -159,6 +164,101 @@ func TestCloneRepo(t *testing.T) {
 	}
 }
 
+func TestCloneRepoSSH(t *testing.T) {
+
```
mafredri, line 168:
```suggestion
```
Maybe we should introduce [gofumpt](https://github.com/mvdan/gofumpt) in this repo. 🤔

mafredri (commented):
> an alternative would be to pass in the SSH key as a base64-encoded file (similar to known_hosts) and clean it up after the build completes.

I think this would be a solid first step at least. We can always iterate/add support for other use-cases later.

envbuilder.go
johnstcn: The default known hosts callback will read the file. I could possibly create a custom known hosts callback that simply iterates over the content of the known_hosts file. WDYT?

envbuilder.go
johnstcn: It will actually refuse to accept that URL, as git URLs are not all valid URLs. 
https://go.dev/play/p/sXvpyxwNLIq

To handle these sorts of URLs, I suppose we could treat all urls without schema prefixes as `ssh://`. WDYT?

BrunoQuaresma: I think it is valid since they don't follow any technical convention format.

envbuilder.go
```diff
@@ -156,6 +157,34 @@ func Run(ctx context.Context, options Options) error {
...
 	}
 
+	gitURLParsed, err := url.Parse(options.GitURL)
+	if err != nil {
+		return fmt.Errorf("invalid git URL: %w", err)
+	}
+	// If we're cloning over SSH, we need a known_hosts file.
+	if gitURLParsed.Scheme == "ssh" {
+		var knownHostsContent []byte
+		if options.GitSSHKnownHostsBase64 != "" {
```
BrunoQuaresma, line 167: Is it sensitive data? Asking because I see it is encoded on base 64 and I'm wondering why.

envbuilder.go
```diff
@@ -156,6 +157,34 @@ func Run(ctx context.Context, options Options) error {
...
+			if kh, err := base64.StdEncoding.DecodeString(options.GitSSHKnownHostsBase64); err != nil {
+				return fmt.Errorf("invalid known_hosts content: %w", err)
```
BrunoQuaresma, line 169: Would make sense to make it more explicit like `invalid Base64 content`? As a user, I would think my `know_hosts` is using some wrong URL or maybe there is a typo error in the config.

git.go
BrunoQuaresma: You are using too much TS Cian 😆

testutil/gittest/gittest.go
```diff
@@ -97,6 +105,97 @@ func NewServer(fs billy.Filesystem) http.Handler {
...
+				if ssh.KeysEqual(pk, key) {
+					return true
+				}
+			}
+			return false
+		}),
+	}
+
+	done := make(chan struct{}, 1)
+	go func() {
```
BrunoQuaresma, line 127: I guess this is some go routine stuff right?

BrunoQuaresma (commented):
Left a few comments but mostly are for my own learning 🙏

envbuilder.go
johnstcn: It's easier to pass data as an environment variable that way; it also follows the existing convention with the docker config JSON.

envbuilder.go
johnstcn: Makes sense!

git.go
```diff
@@ -113,3 +122,55 @@ func CloneRepo(ctx context.Context, opts CloneRepoOptions) (bool, error) {
...
+	k, err := gossh.ParsePrivateKey(buf.Bytes())
+	if err != nil {
+		return nil, fmt.Errorf("parse private key file: %w", err)
+	}
+	return k, nil
+}
+
+// GenerateKnownHosts dials the server located at gitURL and fetches the SSH
+// public keys returned in a format accepted by known_hosts.
+func GenerateKnownHosts(log LoggerFunc, gitURL *url.URL) ([]byte, error) {
```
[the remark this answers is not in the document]
johnstcn: Unfortunately I don't really have a good answer for that, apart from that's what most functions that operate on URLs want, and that's what `url.Parse()` returns. https://pkg.go.dev/net/url#Parse

testutil/gittest/gittest.go
johnstcn: Yeah we don't want `ssh.Serve()` to block execution but we want to wait for it to finish before continuing.

A common pattern here is to use use the fact that recieving on a closed channel will always immediately return the zero value for the channel. So you'l often see stuff like this:

```
done := make(chan struct{})
go func() {
  someBlockingFunction()
  close(done)
}()

<-done // waits for someBlockingFunction to finish
```

envbuilder.go
mafredri: If it's fairly trivial, go for it 👍🏻. If not, let's at least properly document why we do it.

dannykopping (approved):
LGTM! Mostly nits

envbuilder.go
```diff
@@ -156,6 +157,36 @@ func Run(ctx context.Context, options Options) error {
...
+		}
+		// If we're cloning over SSH, we need a known_hosts file.
+		if gitURLParsed.Scheme == "ssh" {
+			var knownHostsContent []byte
+			if options.GitSSHKnownHostsBase64 != "" {
+				if kh, err := base64.StdEncoding.DecodeString(options.GitSSHKnownHostsBase64); err != nil {
+					return fmt.Errorf("invalid known_hosts content: %w", err)
+				} else {
+					knownHostsContent = kh
+				}
```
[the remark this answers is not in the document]
mafredri: Normally I'd agree and prefer to avoiding `else`, but since `kh` is assigned in the `if`-statement, it needs to be in the `else` or then the assignment moved out of `if`. Both are alright approaches IMO.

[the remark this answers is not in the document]
johnstcn: I'm going to simplify this logic a bit.

git_test.go
```diff
@@ -159,6 +164,136 @@ func TestCloneRepo(t *testing.T) {
 	}
 }
 
+func TestCloneRepoSSH(t *testing.T) {
+	t.Parallel()
+
+	t.Run("AuthSuccess", func(t *testing.T) {
```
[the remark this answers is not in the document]
johnstcn: Agreed in principle; but I'd prefer to do this when I have more test cases from which to abstract.

johnstcn:
I've simplifed the logic by removing the need to parse the git URL at all and removing the known_hosts business; if folks need to validate host keys they can simply mount in a known_hosts file and set SSH_KNOWN_HOSTS.
I've also extracted the logic for setting up repo auth into a separate function and added dedicated tests for this.

envbuilder.go
johnstcn: Even more trivial is simply not doing it!

johnstcn:
Bonus points: SSH_AUTH_SOCK also works!

```
docker run -it --rm -e GIT_URL="ssh://git@github.com/johnstcn/test-private-devcontainer" -e SSH_AUTH_SOCK=/tmp/ssh-Vsi58Zsh72/agent.1567 -v /tmp/ssh-Vsi58Zsh72/agent.1567:/tmp/ssh-Vsi58Zsh72/agent.1567 -e INIT_SCRIPT=bash envbuilder:latest
envbuilder - Build development environments from repositories in a container
#1: 📦 Cloning ssh://git@github.com/johnstcn/test-private-devcontainer to /workspaces/test-private-devcontainer...
#1: 🔑 Using SSH authentication!
#1: 🔑 No SSH key found, falling back to agent!
#1: 🔓 SSH_KNOWN_HOSTS not set, accepting all host keys!
#1: 🔑 Got host key: github.com,20.26.156.215 ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBEmKSENjQEezOmxkZMy7opKgwFB9nkt5YRrYMjNuG5N87uRgg6CLrbo5wAdT/y6v0mKV0U2w0WZ2YB/++Tpockg=
```

git_test.go
```diff
@@ -259,39 +260,115 @@ func TestCloneRepoSSH(t *testing.T) {
...
+		auth := envbuilder.SetupRepoAuth(opts)
+		pk, ok := auth.(*gitssh.PublicKeys)
+		require.True(t, ok)
+		require.NotNil(t, pk.Signer)
+		actualSigner, err := gossh.ParsePrivateKey([]byte(testKey))
+		require.NoError(t, err)
+		require.Equal(t, actualSigner, pk.Signer)
+	})
+
+	t.Run("SSH/NoAuthMethods", func(t *testing.T) {
```
dannykopping, line 364: Can you add a case for an invalid scheme?

git_test.go
```diff
@@ -259,39 +260,115 @@ func TestCloneRepoSSH(t *testing.T) {
 	})
 }
 
-func TestParseGitURL(t *testing.T) {
-	t.Parallel()
+// nolint:paralleltest // t.Setenv for SSH_AUTH_SOCK
+func TestSetupRepoAuth(t *testing.T) {
```
dannykopping, line 264: Super!!

dannykopping (approved):
Slick! Love the changes, last comment from me

git.go
```diff
@@ -113,3 +123,131 @@ func CloneRepo(ctx context.Context, opts CloneRepoOptions) (bool, error) {
...
 	return true, nil
 }
+
+// ReadPrivateKey attempts to read an SSH private key from path
+// and returns an ssh.Signer.
+func ReadPrivateKey(path string) (gossh.Signer, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, fmt.Errorf("open private key file: %w", err)
+	}
```
mafredri, line 133:
```suggestion
	}
	defer f.Close()
```

git.go
```diff
@@ -113,3 +123,131 @@ func CloneRepo(ctx context.Context, opts CloneRepoOptions) (bool, error) {
...
+
+	// If no SSH key set, fall back to agent auth.
+	if signer == nil {
+		options.Logger(codersdk.LogLevelError, "#1: 🔑 No SSH key found, falling back to agent!")
+		auth, err := gitssh.NewSSHAgentAuth(authUser)
+		if err != nil {
+			options.Logger(codersdk.LogLevelError, "#1: ❌ Failed to connect to SSH agent: %s", err.Error())
+			return nil // nothing else we can do
+		}
+		if os.Getenv("SSH_KNOWN_HOSTS") == "" {
```
mafredri, line 229: Suggestion: If this was moved higher up, we could avoid duplicating it.

git_test.go
```diff
@@ -159,6 +165,212 @@ func TestCloneRepo(t *testing.T) {
...
+	t.Setenv("SSH_AUTH_SOCK", "")
+	t.Run("Empty", func(t *testing.T) {
+		opts := &envbuilder.Options{
+			Logger: testLog(t),
+		}
+		auth := envbuilder.SetupRepoAuth(opts)
+		require.Nil(t, auth)
+	})
+
+	t.Run("HTTP/NoAuth", func(t *testing.T) {
```
mafredri, line 274: Suggestion: could avoid prefix if these were nested in an otherwise empty `t.Run`. Feel free to disregard if you don't want to do the change. 😄

git.go
```diff
@@ -113,3 +123,131 @@ func CloneRepo(ctx context.Context, opts CloneRepoOptions) (bool, error) {
...
+			Username: options.GitUsername,
+			Password: options.GitPassword,
+		}
+	}
+
+	// Generally git clones over SSH use the 'git' user, but respect
+	// GIT_USERNAME if set.
+	authUser := options.GitUsername
+	if authUser == "" {
+		authUser = "git"
```
mafredri, line 204: Suggestion: consider updating `options.GitUsername` if empty instead, so that we don't need to duplicate this check and can use the option directly.

mafredri (approved):
Looks great!

git_test.go
johnstcn: I think one level of nesting is as deep as we need here in the test :-)

Issue #158, opened 2024-04-29
Keep the comments on the struct and instead go generate the serpent descriptions from those


BrunoQuaresma:
Related to: https://github.com/coder/envbuilder/pull/140
Reference: https://github.com/coder/envbuilder/pull/140#discussion_r1580870391

BrunoQuaresma:
After talking to @Emyrk and @mafredri we decided to move the comments back to the struct and keep them in the CLI Description field too even resulting in duplication because of developing a generator for this use case.

PR #171, opened 2024-05-02
chore: add comments into the Options fields

BrunoQuaresma:
Closes https://github.com/coder/envbuilder/issues/158

### DIFF
--- a/options.go
+++ b/options.go
@@ -13,44 +13,121 @@ type LoggerFunc func(level codersdk.LogLevel, format string, args ...interface{}
 
 // Options contains the configuration for the envbuilder.
 type Options struct {
-	SetupScript          string
-	InitScript           string
-	InitCommand          string
-	InitArgs             string
-	CacheRepo            string
-	BaseImageCacheDir    string
-	LayerCacheDir        string
-	DevcontainerDir      string
+	// SetupScript is the script to run before the init script. It runs as the
+	// root user regardless of the user specified in the devcontainer.json file.
+	// SetupScript is ran as the root user prior to the init script. It is used to
+	// configure envbuilder dynamically during the runtime. e.g. specifying
+	// whether to start systemd or tiny init for PID 1.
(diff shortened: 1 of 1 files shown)


johnstcn (approved):
👍

options.go
```diff
@@ -13,44 +13,121 @@ type LoggerFunc func(level codersdk.LogLevel, format string, args ...interface{}
...
+	LayerCacheDir string
+	// DevcontainerDir is the path to the folder containing the devcontainer.json
+	// file that will be used to build the workspace and can either be an absolute
+	// path or a path relative to the workspace folder. If not provided, defaults
+	// to `.devcontainer`.
+	DevcontainerDir string
+	// DevcontainerJSONPath is the path to a devcontainer.json file that is either
+	// an absolute path or a path relative to DevcontainerDir. This can be used in
+	// cases where one wants to substitute an edited devcontainer.json file for
+	// the one that exists in the repo.
```
mafredri, line 47: Some of these comments look quite different from before the serpent refactor.

```suggestion
	// DevcontainerJSONPath is a path to a devcontainer.json file
	// that is either an absolute path or a path relative to
	// DevcontainerDir. This can be used in cases where one wants
	// to substitute an edited devcontainer.json file for the one
	// that exists in the repo.
	// If neither `DevcontainerDir` nor `DevcontainerJSONPath` is provided,
	// envbuilder will browse following directories to locate it:
	// 1. `.devcontainer/devcontainer.json`
	// 2. `.devcontainer.json`
	// 3. `.devcontainer/<folder>/devcontainer.json`
```

Do you remember why "If neither"-section and the list of files was removed?

There was some other changes too, but this was a big one that stood out to me when I compared. IMO there's some valuable information there that was lost.

BrunoQuaresma: I'm not sure... I may have missed them during copy + paste but I will do a post review to see what we missed.

PR #172, opened 2024-05-03
chore: format go code using gofumpt

BrunoQuaresma:


.github/workflows/ci.yaml
```diff
@@ -77,3 +77,29 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Echo Go Cache Paths
+        id: go-cache-paths
+        run: |
+          echo "GOCACHE=$(go env GOCACHE)" >> ${{ runner.os == 'Windows' && '$env:' || '$' }}GITHUB_OUTPUT
+          echo "GOMODCACHE=$(go env GOMODCACHE)" >> ${{ runner.os == 'Windows' && '$env:' || '$' }}GITHUB_OUTPUT
```
johnstcn, line 92: We don't make windows binaries so the conditional shouldn't be required here

johnstcn (approved):
LGTM, just one nit about the github runner conditionals for windows

.github/workflows/ci.yaml
```diff
@@ -77,3 +77,29 @@ jobs:
...
+        with:
+          path: ${{ steps.go-cache-paths.outputs.GOCACHE }}
+          key: ${{ runner.os }}-go-build-${{ hashFiles('**/go.**', '**.go') }}
+
+      - uses: actions/setup-go@v3
+        with:
+          go-version: "~1.21"
+
+      - name: Check format
+        run: test -z $(go run mvdan.cc/gofumpt -l .)
```
dannykopping, line 105: Do you have an example of a failed CI run?

dannykopping (approved):
LGTM

Makefile
```diff
@@ -1,6 +1,9 @@
 GOARCH := $(shell go env GOARCH)
 PWD=$(shell pwd)
 
+fmt: *.go
```
mafredri, line 4: This will only track the top-level go files, see:

```Makefile
demo: *.go
	ls -1 $^
```

```
~/Work/Code/envbuilder main* ⇣
❯ make demo
envbuilder.go
envbuilder_internal_test.go
envbuilder_test.go
git.go
git_test.go
log.go
log_test.go
options.go
options_test.go
```

vs

```
~/Work/Code/envbuilder main* ⇣ 6s
❯ ls -1 **/*.go
cmd/envbuilder/main.go
devcontainer/devcontainer.go
devcontainer/devcontainer_test.go
devcontainer/features/features.go
devcontainer/features/features_test.go
devcontainer/script.go
devcontainer/script_test.go
envbuilder.go
envbuilder_internal_test.go
envbuilder_test.go
git.go
git_test.go
integration/integration_test.go
log.go
log_test.go
options.go
options_test.go
scripts/docsgen/main.go
testutil/gittest/gittest.go
testutil/registrytest/registrytest.go
```

.github/workflows/ci.yaml
mafredri: Since this is just a boolean check, perhaps it can be more helpful if we show what went wrong? How about:

```shell
list=$(go run mvdan.cc/gofumpt -l .)
if [[ -n $list ]]; then
	echo -e "error: The following files have changes:\n\n${list}\n\nDiff:\n\n"
	go run mvdan.cc/gofumpt -d .
	exit 1
fi
```

.github/workflows/ci.yaml
```diff
@@ -77,3 +77,29 @@ jobs:
...
+          echo "GOCACHE=$(go env GOCACHE)" >> ${{ runner.os == 'Windows' && '$env:' || '$' }}GITHUB_OUTPUT
+          echo "GOMODCACHE=$(go env GOMODCACHE)" >> ${{ runner.os == 'Windows' && '$env:' || '$' }}GITHUB_OUTPUT
+
+      - name: Go Build Cache
+        uses: actions/cache@v3
+        with:
+          path: ${{ steps.go-cache-paths.outputs.GOCACHE }}
+          key: ${{ runner.os }}-go-build-${{ hashFiles('**/go.**', '**.go') }}
+
+      - uses: actions/setup-go@v3
```
matifali, line 100:
```suggestion
      - uses: actions/setup-go@v5
```
This has built in cache so we should use that and make CI simple

.github/workflows/ci.yaml
```diff
@@ -77,3 +77,29 @@ jobs:
...
+        id: go-cache-paths
+        run: |
+          echo "GOCACHE=$(go env GOCACHE)" >> ${{ runner.os == 'Windows' && '$env:' || '$' }}GITHUB_OUTPUT
+          echo "GOMODCACHE=$(go env GOMODCACHE)" >> ${{ runner.os == 'Windows' && '$env:' || '$' }}GITHUB_OUTPUT
+
+      - name: Go Build Cache
+        uses: actions/cache@v3
+        with:
+          path: ${{ steps.go-cache-paths.outputs.GOCACHE }}
+          key: ${{ runner.os }}-go-build-${{ hashFiles('**/go.**', '**.go') }}
```
matifali, line 98:
```suggestion
```

.github/workflows/ci.yaml
BrunoQuaresma: Using @mafredri suggestion the output would look like this:
```
error: The following files have changes:

envbuilder.go

Diff:


diff envbuilder.go.orig envbuilder.go
--- envbuilder.go.orig
+++ envbuilder.go
@@ -151,7 +151,7 @@
                if err != nil {
                        return fmt.Errorf("parse docker config: %w", err)
                }
-               err = os.WriteFile(filepath.Join(MagicDir, "config.json"), decoded, 0644)
+               err = os.WriteFile(filepath.Join(MagicDir, "config.json"), decoded, 0o644)
                if err != nil {
                        return fmt.Errorf("write docker config: %w", err)
                }
```

.github/workflows/ci.yaml
```diff
@@ -77,3 +77,29 @@ jobs:
 
       - name: Check for unstaged files
         run: git diff --exit-code
+  fmt:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
matifali, line 86:
```suggestion
```
This will wail the workflow if we run it manually using `workflow_dispatch`.
From docs:
>  The branch, tag or SHA to checkout. When checking out the repository that
   triggered a workflow, this defaults to the reference or SHA for that event.
   Otherwise, uses the default branch.
   ref: ''

matifali (changes requested):

.github/workflows/ci.yaml
mafredri: Looks like I added one \n too much 😅, and maybe the "Diff:" is superfluous. Thanks for trying it out @BrunoQuaresma!

BrunoQuaresma:
@matifali should we make test, docs, and fmt required?

matifali:
@BrunoQuaresma Yes I think. Someone with admin rights to the repo can mark jobs as required.

.github/workflows/ci.yaml
```diff
@@ -77,3 +67,15 @@ jobs:
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version: "~1.21"
+
+      - name: Check format
+        run: bash ./scripts/check_fmt.sh
```
mafredri, line 81: It's better to not call `bash` here and execute the script directly (make sure script is `chmod +x`). Otherwise I believe this will override the `#!` header in the script file meaning it _could_ run in an incompatible shell.

Makefile
```diff
@@ -1,6 +1,9 @@
 GOARCH := $(shell go env GOARCH)
 PWD=$(shell pwd)
 
+fmt: **/*.go
```
mafredri, line 4: Did you confirm that this works? I've never seen this syntax in makefiles before.

scripts/check_fmt.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+list="$(go run mvdan.cc/gofumpt@v0.6.0 -l .)"
+if [[ -n $list ]]; then
+  echo -e "error: The following files have changes:\n\n${list}\n\nDiff:\n\n"
```
mafredri, line 5:
```suggestion
  echo -n -e "error: The following files have changes:\n\n${list}\n\nDiff:\n\n"
```

.github/workflows/ci.yaml
BrunoQuaresma: Good call, I already had set the permissions with `chmod +x`

BrunoQuaresma:
@mafredri follow-ups here https://github.com/coder/envbuilder/pull/179

PR #173, opened 2024-05-03
fix: use serpent.String instead of serpent.URL for CODER_AGENT_URL

johnstcn:
Workaround for https://github.com/coder/serpent/issues/14

### DIFF
--- a/cmd/envbuilder/main.go
+++ b/cmd/envbuilder/main.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"net/url"
 	"os"
 	"slices"
 	"strings"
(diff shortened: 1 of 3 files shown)


options.go
```diff
@@ -132,7 +131,7 @@ type Options struct {
 	Filesystem billy.Filesystem
 	// These options are specifically used when envbuilder is invoked as part of a
 	// Coder workspace.
-	CoderAgentURL *url.URL
+	CoderAgentURL string
```
mafredri, line 134: Should we add this in?

```suggestion
	// Revert to `*url.URL` once https://github.com/coder/serpent/issues/14 is fixed.
	CoderAgentURL string
```

cmd/envbuilder/main.go
```diff
@@ -31,10 +32,14 @@ func main() {
...
 			var sendLogs func(ctx context.Context, log ...agentsdk.Log) error
 			if options.CoderAgentToken != "" {
-				if options.CoderAgentURL == nil {
+				if options.CoderAgentURL == "" {
 					return errors.New("CODER_AGENT_URL must be set if CODER_AGENT_TOKEN is set")
 				}
-				client := agentsdk.New(options.CoderAgentURL)
+				u, err := url.Parse(options.CoderAgentURL)
+				if err != nil {
+					return fmt.Errorf("invalid value for CODER_AGENT_URL: %w", err)
```
mafredri, line 40: Suggestion: more descriptive?

```suggestion
					return fmt.Errorf("unable to parse CODER_AGENT_URL as URL: %w", err)
```

mafredri (approved):

PR #174, opened 2024-05-03
feat: fetch SSH key from Coder if required

johnstcn:
Relates to https://github.com/coder/envbuilder/issues/31

Depends on https://github.com/coder/envbuilder/pull/173

Modifies the SSH logic introduced in https://github.com/coder/envbuilder/pull/170 with an extra step to fetch the SSH key from Coder.

There is currently a race condition between envbuilder starting and the workspace build being marked as 'completed'. To work around this, I added some backoff to fetching the SSH key using the agent token.

Demo:

https://github.com/coder/envbuilder/assets/4949514/986fc223-c50c-47dc-afc1-a251268cf0a0

### DIFF
--- a/git.go
+++ b/git.go
@@ -176,6 +176,13 @@ func LogHostKeyCallback(log LoggerFunc) gossh.HostKeyCallback {
 // If SSH_PRIVATE_KEY_PATH is set, an SSH private key will be read from
 // that path and the SSH auth method will be configured with that key.
 //
+// If no SSH_PRIVATE_KEY_PATH is set, but CODER_AGENT_URL and CODER_AGENT_TOKEN
+// are both specified, envbuilder will attempt to fetch the corresponding
+// Git SSH key for the user.
+//
+// Otherwise, SSH authentication will fall back to SSH_AUTH_SOCK, in which
+// case SSH_AUTH_SOCK must be set to the path of a listening SSH agent socket.
+//
 // If SSH_KNOWN_HOSTS is not set, the SSH auth method will be configured
 // to accept and log all host keys. Otherwise, host key checking will be
 // performed as usual.
--- a/git.go
+++ b/git.go
@@ -7,11 +7,13 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"net/http"
 	"net/url"
 	"os"
 	"strings"
 	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/coder/coder/v2/codersdk"
 	"github.com/coder/coder/v2/codersdk/agentsdk"
 	"github.com/go-git/go-billy/v5"
@@ -186,7 +188,7 @@ func LogHostKeyCallback(log LoggerFunc) gossh.HostKeyCallback {
 // If SSH_KNOWN_HOSTS is not set, the SSH auth method will be configured
 // to accept and log all host keys. Otherwise, host key checking will be
 // performed as usual.
-func SetupRepoAuth(options *Options) transport.AuthMethod {
+func SetupRepoAuth(ctx context.Context, options *Options) transport.AuthMethod {
 	if options.GitURL == "" {
 		options.Logger(codersdk.LogLevelInfo, "#1: ❔ No Git URL supplied!")
 		return nil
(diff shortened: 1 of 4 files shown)


git.go
```diff
@@ -251,3 +277,63 @@ func SetupRepoAuth(options *Options) transport.AuthMethod {
...
+		s, err := FetchCoderSSHKey(ctx, coderURL, agentToken)
+		if err != nil {
+			var sdkErr *codersdk.Error
+			if errors.As(err, &sdkErr) && sdkErr.StatusCode() == http.StatusUnauthorized {
+				// Retry, as this may just mean that the workspace build has not yet
+				// completed.
+				log(codersdk.LogLevelInfo, "#1: 🕐 Backing off as the workspace build has not yet completed...")
+				return err
+			}
+			close(signerChan)
```
johnstcn, line 305: review: I'm open to adding more cases here, but keeping the retry scope small for now.

git.go
```diff
@@ -251,3 +277,63 @@ func SetupRepoAuth(options *Options) transport.AuthMethod {
...
+	}
+	return signer, nil
+}
+
+// keyFingerprint returns the md5 checksum of the public key of signer.
+func keyFingerprint(s gossh.Signer) string {
+	h := md5.New()
+	h.Write(s.PublicKey().Marshal())
+	return fmt.Sprintf("%x", h.Sum(nil))
+}
```
johnstcn, line 339: This is an unrelated change, but I figured it would be nice to see

git.go
```diff
@@ -251,3 +277,63 @@ func SetupRepoAuth(options *Options) transport.AuthMethod {
...
+// It will retry for up to 1 minute with exponential backoff.
+// Any other error is considered a permanent failure.
+func FetchCoderSSHKeyRetry(ctx context.Context, log LoggerFunc, coderURL, agentToken string) (gossh.Signer, error) {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	signerChan := make(chan gossh.Signer, 1)
+	eb := backoff.NewExponentialBackOff()
+	eb.MaxElapsedTime = 0
+	eb.MaxInterval = time.Minute
```
johnstcn, line 293: review: should this maybe be a configurable knob?

matifali:
Is it simular to #92?

git.go
```diff
@@ -207,14 +218,29 @@ func SetupRepoAuth(options *Options) transport.AuthMethod {
...
+		options.Logger(codersdk.LogLevelInfo, "#1: 🔑 Fetching key from %s!", options.CoderAgentURL)
+		fetchCtx, cancel := context.WithCancel(ctx)
+		defer cancel()
+		s, err := FetchCoderSSHKeyRetry(fetchCtx, options.Logger, options.CoderAgentURL, options.CoderAgentToken)
+		if err == nil {
+			signer = s
+			options.Logger(codersdk.LogLevelInfo, "#1: 🔑 Fetched %s key %s !", signer.PublicKey().Type(), keyFingerprint(signer)[:8])
+		} else {
+			options.Logger(codersdk.LogLevelInfo, "#1: ❌ Failed to fetch SSH key from %s: %w", options.CoderAgentURL, err)
 		}
```
kylecarbs, line 244: Deeply integrating the hot-path code with Coder feels incorrect to me.

I don't believe the `envbuilder` package should have any knowledge of the Coder agent or Coder URL. Prior to the CLI change, it didn't. The command-line simply logged to Coder as a sink if those vars were provided, and I think implementing this similarly abstractly would be an improvement.

We want people outside of the Coder ecosystem to use envbuilder. By integrating the codebase so heavily into Coder, we alienate that possibility.

kylecarbs: My preference would be to allow passing SSH options to envbuilder, and also not provide Coder options directly to envbuilder.

johnstcn: How about a pre-init hook command?

nwrkbiz:
Awesome, this is was I am waiting for since a while 🙏🏻

git.go
mtojek: Adopting "lifecycle hooks" is a reasonable concept, we may argue where they should be scripts or just shell commands. Maybe we can implement them _similarly_ to [Git Hooks](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks)?

mtojek (commented):
I'm happy with exploring the idea for hooks unless https://github.com/coder/envbuilder/issues/31 is bleeding hard (urgent).

git.go
mafredri: Any thoughts as to what kind of hook, except pre-clone, we might need in envbuilder? Running scripts is generally already addressable by `devcontainer.json`.

mafredri: If we added a new data source to the coder provider, I believe we address this similarly to what @kylecarbs suggested:

```
container {
  env = {
    GIT_SSH_KEY: basy64(data.coder_workspace.owner_ssh_key)
  }
}
```

kylecarbs: That sounds like a wonderful way to do it!

matifali: This is probably out of scope but I want to share my thoughts,

As the ssh key is per user and not per workspace we should probably add/refactor a new data source `coder_user`

```
container {
  env = {
    GIT_SSH_KEY: basy64(data.coder_owner.private_ssh_key)
  }
}
```

Others properties we can add or move from `coder_workspace` can be
1. `coder_workspace.owner` -> `coder_user.name`
2. `coder_workspace.owner_name` -> `coder_user.full_name`
3. `coder_workspace.owner_emial` -> `coder_user.email`
4. Add `coder_user.private_ssh_key`
5. Add `coder_user.public_ssh_key`
6. Add `coder_user.groups` (A list of all groups if applicable)

cc: @bpmct

bpmct: Yeah, I think you're right that another data source for the coder_user is good, but shouldn't block this effort.

bpmct: Mentioning that in https://github.com/coder/terraform-provider-coder/issues/217

johnstcn: > If we added a new data source to the coder provider, I believe we can address this similarly to what @kylecarbs suggested:
> 
> ```terraform
> container {
>   env = {
>     GIT_SSH_KEY: basy64(data.coder_workspace.owner_ssh_key)
>   }
> }
> ```

👍 That's certainly a valid path forward.

Based on what was implemented in #170, there would be a few extra steps:
- Create a secret containing the user's private key
- Mount that secret into the container under a well-known path
- Reference that path in the container's environment

Will open a separate PR for this.

matifali: Already moved. See
https://github.com/coder/terraform-provider-coder/issues/219

johnstcn:
Thanks @mafredri @matifali for your excellent input and comments.

The overwhelming consensus here is to avoid a direct integration and instead allow injecting a Coder user's SSH private key into the envbuilder container. There should be no further changes in envbuilder required to support this. Closing this PR out in favour of https://github.com/coder/terraform-provider-coder/issues/219

Issue #166, opened 2024-05-01
(breaking) Prefix all envbuilder environment variables with ENVBUILDER_


johnstcn:
For consistency with `coder/coder`, we should prefix all environment variables consumed by envbuilder with the string `ENVBUILDER_`.

This will be a breaking change and must be done before any guarantee of future stability.

bpmct:
Totally makes sense. If we were to support https://github.com/coder/envbuilder/issues/127, how do you imagine that would work? `ENVBUILDER_KANIKO_*`?

johnstcn:
Envbuilder currently imports kaniko as a library, so I don't believe this would impact how kaniko reads its environment variables.

bpmct:
So if we wanted to support all (or most) kaniko options, we would rewrite instead of passing through?

johnstcn:
`KANIKO_DIR` is just looked up on-demand at present so we do not need to do anything for that env.

Other kaniko options are checked in its root command (for example, `KANIKO_REGISTRY_MIRROR`). We currently have to check for the presence of this environment variable and pass it in as required. Other similar environment variables include `KANIKO_REGISTRY_MAP` and `KANIKO_NO_PUSH`. I don't think we would want to prefix in this case.

BrunoQuaresma:
After discussing this on Coconut standup we want to:
- Add env variables prefix option to the coder/serpent and after support this on the envbuilder 
- The "prefix env variable" should still support the env variables without the prefix so it does not break old consumers

matifali:
> The "prefix env variable" should still support the env variables without the prefix so it does not break old consumers

We should only support old format until we do 1.0 we should break when we release 1.0.

BrunoQuaresma:
Thinking more about how we want to make it, I wonder if we could make `coder/serpent` to accept `Env` as an `[]string` instead. IMO, it makes the code more clear and direct about the usage. Another option would be to `Env` as `any` and accept `string` or `string[]`. 

```go
const prefix = "CODER_"

return serpent.OptionSet{
  {
    Flag:  "setup-script",
    Env:   []string{"SETUP_SCRIPT", prefix + "SETUP_SCRIPT"},
    Value: serpent.StringOf(&o.SetupScript),
    Description: "...",
  },
}
```

Wdyt?

matifali:
[...]

PR #180, opened 2024-05-10
feat: prefix env variables with ENVBUILDER

BrunoQuaresma:
Closes https://github.com/coder/envbuilder/issues/166

### DIFF
--- a/options.go
+++ b/options.go
@@ -140,14 +140,14 @@ type Options struct {
 	CoderAgentSubsystem []string
 }
 
-const optionPrefix = "ENVBUILDER_"
+const envPrefix = "ENVBUILDER_"
 
 // Generate CLI options for the envbuilder command.
 func (o *Options) CLI() serpent.OptionSet {
 	options := serpent.OptionSet{
 		{
 			Flag:  "setup-script",
-			Env:   WithOptionPrefix("SETUP_SCRIPT"),
+			Env:   WithEnvPrefix("SETUP_SCRIPT"),
 			Value: serpent.StringOf(&o.SetupScript),
 			Description: "The script to run before the init script. It runs as " +
 				"the root user regardless of the user specified in the devcontainer.json " +
@@ -157,21 +157,21 @@ func (o *Options) CLI() serpent.OptionSet {
 		},
 		{
 			Flag:        "init-script",
-			Env:         WithOptionPrefix("INIT_SCRIPT"),
+			Env:         WithEnvPrefix("INIT_SCRIPT"),
 			Default:     "sleep infinity",
 			Value:       serpent.StringOf(&o.InitScript),
 			Description: "The script to run to initialize the workspace.",
 		},
 		{
 			Flag:        "init-command",
-			Env:         WithOptionPrefix("INIT_COMMAND"),
+			Env:         WithEnvPrefix("INIT_COMMAND"),
 			Default:     "/bin/sh",
 			Value:       serpent.StringOf(&o.InitCommand),
 			Description: "The command to run to initialize the workspace.",
 		},
 		{
 			Flag:  "init-args",
-			Env:   WithOptionPrefix("INIT_ARGS"),
+			Env:   WithEnvPrefix("INIT_ARGS"),
 			Value: serpent.StringOf(&o.InitArgs),
 			Description: "The arguments to pass to the init command. They are " +
 				"split according to /bin/sh rules with " +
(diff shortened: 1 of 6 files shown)


options.go
```diff
@@ -395,15 +397,40 @@ func (o *Options) CLI() serpent.OptionSet {
 				"The envbuilder subsystem is always included.",
 		},
 	}
+
+	// Add options without the prefix for backward compatibility. These options
+	// are marked as deprecated and will be removed in future versions. Note:
+	// Future versions will require the 'ENVBUILDER' prefix for default
```
mtojek, line 403: nit: `ENVBUILDER_ ` (underscore)

options.go
```diff
@@ -395,15 +397,40 @@ func (o *Options) CLI() serpent.OptionSet {
...
+	// environment variables.
+	for _, o := range set {
+		if strings.HasPrefix(o.Env, prefix) {
+			prevOption := o
+			prevOption.Env = strings.TrimPrefix(o.Env, prefix)
+			prevOption.UseInstead = []serpent.Option{o}
+			prevOption.Hidden = true
+			set = append(set, prevOption)
```
mtojek, line 411: that is risky, you're modifying the object you're looping. Let's make this idempotent: 

`opts := supportLegacyEnvWithoutPrefixes(opts)`

options.go
```diff
@@ -142,10 +142,12 @@ type Options struct {
 
 // Generate CLI options for the envbuilder command.
 func (o *Options) CLI() serpent.OptionSet {
-	return serpent.OptionSet{
+	const prefix = "ENVBUILDER_"
+
+	set := serpent.OptionSet{
```
mtojek, line 147: `set` may not be a candidate for the name, `opts` or `options`?

options.go
```diff
@@ -142,10 +142,12 @@ type Options struct {
...
 		{
 			Flag:  "setup-script",
-			Env:   "SETUP_SCRIPT",
+			Env:   prefix + "SETUP_SCRIPT",
```
mtojek, line 150: hint: I used to add a helper function `withPrefix("SETUP_SCRIPT)` to have a single place where I concatenate strings.

options.go
```diff
@@ -395,15 +397,40 @@ func (o *Options) CLI() serpent.OptionSet {
...
 
 func (o *Options) Markdown() string {
-	cliOptions := o.CLI()
+	var visibleOpts serpent.OptionSet
+
+	// We don't want to include deprecated opts in the README.
+	for _, opt := range o.CLI() {
+		isDeprecated := len(opt.UseInstead) > 0
+		if !isDeprecated {
+			visibleOpts = append(visibleOpts, opt)
```
mtojek, line 425: naming: `visibleOpts` could be rephrased to `activeOptions` or `validOptions`.

options.go
```diff
@@ -395,15 +397,40 @@ func (o *Options) CLI() serpent.OptionSet {
...
+
+	return set
 }
 
 func (o *Options) Markdown() string {
-	cliOptions := o.CLI()
+	var visibleOpts serpent.OptionSet
+
+	// We don't want to include deprecated opts in the README.
+	for _, opt := range o.CLI() {
```
mtojek, line 422: Technically speaking, you can move this for-range loop to another function and have:

`cliOptions := skipDeprecatedOptions(cliOptions)`

mtojek (commented):
Do you think we need any unit tests here?

options.go
BrunoQuaresma: Not sure what this means, when doing `prevOption := o` I'm passing only the value and not the `o` reference so I'm not sure what is the risk. Ps: It is a learning note.

options.go
BrunoQuaresma: Learning question: What value do you see in having a single place to concatenate strings?

options.go
mtojek: Modifying a slice (set) while iterating over it can lead to unpredictable behavior. It's generally not recommended to modify the collection you're iterating over. This can lead to unexpected results or even runtime errors.

options.go
mtojek: You can prevent developer errors in the future. Let's say, a developer would like to search&replace `prefix +` with `prefix2 +`. Now, there are 10+ places to change = easy to miss one. Having `withPrefix(env string) { }` function, you have only place to do this :)

BrunoQuaresma:
> Do you think we need any unit tests here?

Yes, I was waiting for some review to see if I was in the right direction.  👍

mtojek:
Awesome! Looking forward to see some tests 👍

options_test.go
```diff
@@ -15,54 +15,66 @@ import (
...
+		legacyEnvValue := "./setup-legacy-script.sh"
+		t.Setenv("SETUP_SCRIPT", legacyEnvValue)
+		o := runCLI()
+		require.Equal(t, o.SetupScript, legacyEnvValue)
+
+		envValue := "./setup-script.sh"
+		t.Setenv(envbuilder.WithEnvPrefix("SETUP_SCRIPT"), envValue)
+		o = runCLI()
+		require.Equal(t, o.SetupScript, envValue)
+	})
```
BrunoQuaresma, line 77: @mtojek Idk what is the best way to test the legacy support for the options but I just thought about getting one of the options and testing the values with the prefix and without it. Wdyt?

BrunoQuaresma: @johnstcn answered my question here https://github.com/coder/envbuilder/pull/180#discussion_r1596967496.

options_test.go
```diff
@@ -15,54 +15,66 @@ import (
...
 
 		t.Run("empty", func(t *testing.T) {
-			t.Setenv("GIT_CLONE_SINGLE_BRANCH", "")
+			t.Setenv(envbuilder.WithEnvPrefix("GIT_CLONE_SINGLE_BRANCH"), "")
 			o := runCLI()
 			require.False(t, o.GitCloneSingleBranch)
 		})
 	})
+
+	t.Run("legacy", func(t *testing.T) {
```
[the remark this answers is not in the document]
BrunoQuaresma: Hm... ok, let's do it.

johnstcn (approved):
Nice thorough tests :+1:

mtojek (approved):
👍

mtojek:
just make sure that CI is happy

PR #183, opened 2024-05-13
feat: temporarily remount read-only mounts before cleaning container fs

johnstcn:
This PR is a continuation of @janLo's original PR https://github.com/coder/envbuilder/pull/51 with some additional changes:

- Brings original PR up to date with latest main
- Moves the temp-remount logic into a separate package
- Refactors for ease of testing and adds unit tests

Manual verification:

- (Existing behaviour): On Docker runtime, dockerd fails to start after running:
  ```shell
    docker run -it --rm \     
      -e GIT_URL=https://github.com/coder/envbuilder \                     
      -e INIT_SCRIPT=bash \
      ghcr.io/coder/envbuilder:latest
  ```

- (New behaviour): Using sysbox-runc runtime\*, dockerd starts succesfully:
    ```shell
    docker run -it --rm \     
      -e GIT_URL=https://github.com/coder/envbuilder \                     
      -e INIT_SCRIPT=bash \
      envbuilder:latest # local built image
  ```

- Existing sample devcontainer builds successfully on Docker runtime:
  ```shell
    docker run -it --rm \     
      -e GIT_URL=https://github.com/coder/envbuilder-starter-devcontainer \                     
      -e INIT_SCRIPT=bash \
      envbuilder:latest # local built image
  ```

\* Lima config used: https://gist.github.com/johnstcn/b9d6854b3aedb551af1a7588e2a5af94


internal/util/remount.go
```diff
@@ -0,0 +1,123 @@
...
+				logf(codersdk.LogLevelTrace, "skip mount %s under ignored prefix %s", mountInfo.MountPoint, prefix)
+				continue outer
+			}
+		}
+
+		src := mountInfo.MountPoint
+		tgt := filepath.Join("/", dest, src)
+		err := m.MkdirAll(tgt, 0o750)
+		if err != nil {
+			return remount, fmt.Errorf("create temp mountpoint %s: %w", dest, err)
```
mafredri, line 74: Should we return `nil` for remount here? And should we perhaps restore the mounts on error?

envbuilder.go
```diff
@@ -394,6 +396,14 @@ func Run(ctx context.Context, options Options) error {
 		})
 	}
 
+	// temp move of all ro mounts
+	tempRemountDest := filepath.Join("/", MagicDir)
+	ignorePrefixes := []string{tempRemountDest, "/proc", "/sys"}
```
dannykopping, line 401: Should these be configurable?
Will we only ever support UNIX-based builds?

internal/util/remount.go
```diff
@@ -0,0 +1,123 @@
...
+
+// TempRemount iterates through all read-only mounted filesystems, bind-mounts them at dest,
+// and unmounts them from their original source. All mount points underneath ignorePrefixes
+// will not be touched.
+//
+// It is the responsibility of the caller to call the returned function
+// to restore the original mount points. If an error is encountered while attempting to perform
+// the operation, calling the returned remount function will make a best-effort attempt to
+// restore the original state.
+func TempRemount(logf func(codersdk.LogLevel, string, ...any), dest string, ignorePrefixes ...string) (remount func() error, err error,
```
mafredri, line 22:
```suggestion
func TempRemount(logf func(codersdk.LogLevel, string, ...any), dest string, ignorePrefixes ...string) (restore func() error, err error,
```

envbuilder.go
```diff
@@ -394,6 +396,14 @@ func Run(ctx context.Context, options Options) error {
 		})
 	}
 
+	// temp move of all ro mounts
+	tempRemountDest := filepath.Join("/", MagicDir)
```
dannykopping, line 400: Not introduced by your PR but I want to point out that `MagicDir` is kinda unhelpful as a name; perhaps `EnvbuilderWorkingDir` would be more explicit?

internal/util/remount.go
```diff
@@ -0,0 +1,123 @@
...
+) {
+	return tempRemount(&realMounter{}, logf, dest, ignorePrefixes...)
```
dannykopping, line 24: Why the indirection here?

internal/util/remount.go
```diff
@@ -0,0 +1,123 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+	"syscall"
+
+	"github.com/coder/coder/v2/codersdk"
+	"github.com/prometheus/procfs"
+)
+
+// TempRemount iterates through all read-only mounted filesystems, bind-mounts them at dest,
```
mafredri, line 14: This is an interesting question, should we or should we not limit this to read-only mounted filesystems? If we think about how a docker container is run with `-v /path:/path`, the contents of the docker container can never populate the mounted `/path`.

internal/util/remount.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+
+		err = m.Mount(src, tgt, "bind", syscall.MS_BIND, "")
+		if err != nil {
+			return remount, fmt.Errorf("bind mount %s => %s: %s", src, dest, err.Error())
+		}
+		err = m.Unmount(src, 0)
+		if err != nil {
+			return remount, fmt.Errorf("temp unmount src %s: %s", src, err.Error())
+		}
```
dannykopping, line 84: Seems like duplication; work extraction to a method?

envbuilder.go
mafredri, line 400: I wonder if we should make this `/[MagicDir]/mnt`

envbuilder.go
```diff
@@ -540,6 +550,11 @@ func Run(ctx context.Context, options Options) error {
 		closeAfterBuild()
 	}
 
+	if err := remount(); err != nil {
+		// Don't fail at this point as it may be useful to be able to inspect the build.
+		options.Logger(codersdk.LogLevelError, "recreate mountpoint: %w", err)
```
mafredri, line 555: Should we error out here?

envbuilder.go
```diff
@@ -394,6 +396,14 @@ func Run(ctx context.Context, options Options) error {
 		})
 	}
 
+	// temp move of all ro mounts
+	tempRemountDest := filepath.Join("/", MagicDir)
+	ignorePrefixes := []string{tempRemountDest, "/proc", "/sys"}
+	remount, err := ebutil.TempRemount(options.Logger, tempRemountDest, ignorePrefixes...)
+	if err != nil {
+		return fmt.Errorf("temp remount: %w", err)
+	}
```
mafredri, line 405: I was going to suggest a defer here, but perhaps it's not so important that we restore the mounts on error.

dannykopping (approved):
LGTM, minor Qs and nits

internal/util/remount.go
mtojek, line 14: Is there a strong point why we should keep it in `internal`?

internal/util/remount.go
```diff
@@ -0,0 +1,123 @@
...
+	"strings"
+	"syscall"
+
+	"github.com/coder/coder/v2/codersdk"
+	"github.com/prometheus/procfs"
+)
+
+// TempRemount iterates through all read-only mounted filesystems, bind-mounts them at dest,
+// and unmounts them from their original source. All mount points underneath ignorePrefixes
+// will not be touched.
```
mtojek, line 16: You may want to add some context in the comment about why/when it is needed.

internal/util/mock_mounter_test.go
```diff
@@ -0,0 +1,98 @@
+// Code generated by MockGen. DO NOT EDIT.
+// Source: ./internal/util/remount.go
+//
+// Generated by this command:
+//
+//	mockgen -source ./internal/util/remount.go -package ebutil mounter
```
mtojek, line 6: auto mockgen in Makefile as a side issue?

internal/util/mock_mounter_test.go
```diff
@@ -0,0 +1,98 @@
+// Code generated by MockGen. DO NOT EDIT.
```
mtojek, line 1: Did you use the mock to have it tested without the OS? Maybe it is more natural to write a real Linux-only test.

envbuilder.go
johnstcn: Kaniko is currently Linux-only to the best of my knowledge.

envbuilder.go
johnstcn: I can put the logic inside a `sync.Once` and then we can just `defer restore()`

envbuilder.go
johnstcn: Honestly I'm in two minds about it, but if we fail to remount then it's unlikely that building the image will succeed. Erroring out here is probably better.

internal/util/remount.go
johnstcn: Only reason is so that it doesn't get imported by anyone else.

internal/util/remount.go
johnstcn: I don't want the caller in the real world to have to care about what `mounter` implementaion they use.

internal/util/remount.go
johnstcn: I'd prefer to leave the responsibility to the caller. Is there a strong reason to do it ourselves?

envbuilder.go
johnstcn: Not a bad idea 👍

johnstcn: https://github.com/coder/envbuilder/issues/184

internal/util/mock_mounter_test.go
johnstcn: Tests that modify the underlying system are extremely obnoxious and tend to be flaky. I'd prefer to just keep the mock test for now. This still gets tested by the integration tests.

envbuilder.go
kylecarbs: I'm not sure prefixing with envbuilder is that clarifying, just considering it's in the envbuilder repo.

I'm happy to change it from MagicDir though.

internal/util/remount.go
mtojek: I'm always in favor of keeping things open, but I won't strongly oppose here :)

internal/util/mock_mounter_test.go
mtojek: > This still gets tested by the integration tests

Ok, that answers my concerns 👍

mtojek (approved):

envbuilder.go
dannykopping: `WorkingDir` SGTM too 👍

internal/util/remount.go
dannykopping: Whoops, missed that 👍

internal/util/remount.go
johnstcn: Yeah this would actually make sense given the below behaviour:

```
$ echo 'important stuff' >> /opt/important-stuff/tps-reports.txt
$ ls -l /opt/important-stuff/tps-reports.txt 
-rw-r--r-- 1 cian 16 May 13 21:12 /opt/important-stuff/tps-reports.txt
$ docker run -it --rm \
    -v /opt/important-stuff:/opt/important-stuff \
    -e GIT_URL=https://github.com/coder/envbuilder-starter-devcontainer \
    -e INIT_SCRIPT=bash \
    envbuilder:latest
[...]
devcontainer # ls -l /opt/important-stuff/
total 0
devcontainer # exit
$ cat /opt/important-stuff/tps-reports.txt
important stuff
```

johnstcn: I might merge this as-is and investigate rw mounts in a follow up, if that's OK?

johnstcn: https://github.com/coder/envbuilder/issues/187

internal/util/remount.go
mafredri: I'd say the main reason is meeting expectations, i.e. returning an err _and_ a non-nil value that is actionable can lead to some bad assumptions, especially if the behavior changes at some point.

It's OK to leave as-is, though.

internal/ebutil/remount.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		return func() error { return nil }, fmt.Errorf("get mounts: %w", err)
+	}
+
+	// temp move of all ro mounts
+	mounts := map[string]string{}
+	// closer to attempt to restore original mount points
+	var restoreOnce sync.Once
+	restore = func() error {
+		var merr error
```
mafredri, line 52: Should we move this outside the func so we can echo `merr` on consecutive calls? Although probably best to not have the error logged twice.

internal/ebutil/remount.go
```diff
@@ -0,0 +1,134 @@
...
+
+		for _, prefix := range ignorePrefixes {
+			if strings.HasPrefix(mountInfo.MountPoint, prefix) {
+				logf(codersdk.LogLevelTrace, "skip mount %s under ignored prefix %s", mountInfo.MountPoint, prefix)
+				continue outer
+			}
+		}
+
+		src := mountInfo.MountPoint
+		dest := filepath.Join("/", base, src)
```
mafredri, line 79: I think it's best to leave this to the caller:

```suggestion
		dest := filepath.Join(base, src)
```

mafredri (approved):
A few minor observations, but otherwise LGTM, nice work!

PR #191, opened 2024-05-15
chore: add docs re docker inside envbuilder-built-envs

johnstcn:


docs/docker.md
```diff
@@ -0,0 +1,108 @@
...
+you are experimenting on your own workstation).
+
+Example:
+
+```shell
+docker run -it --rm \
+    -v /tmp/envbuilder:/workspaces \
+    -e ENVBUILDER_GIT_URL=https://github.com/coder/envbuilder \
+    -e ENVBUILDER_DEVCONTAINER_DIR=/workspaces/envbuilder/examples/docker/01_dood \
+    -e ENVBUILDER_INIT_SCRIPT=bash \
```
mafredri, line 25: Missing socket.

docs/docker.md
```diff
@@ -0,0 +1,108 @@
...
+docker run -it --rm \
+    --privileged \
+    -v /tmp/envbuilder:/workspaces \
+    -e ENVBUILDER_GIT_URL=https://github.com/coder/envbuilder \
+    -e ENVBUILDER_DEVCONTAINER_DIR=/workspaces/envbuilder/examples/docker/02_dind \
+    -e ENVBUILDER_INIT_SCRIPT=/entrypoint.sh \
+    ghcr.io/coder/envbuilder:latest
+```
+
+This can also be accomplished using the [`docker-in-docker` Devcontainer feature](https://github.com/devcontainers/features/tree/main/src/docker-in-docker).
```
mafredri, line 56: Just off the top of my head, this won't work OOTB, envbuilder doesn't run the entrypoint defined here:

https://github.com/devcontainers/features/blob/67c10a660868260c284b02f2878dcfdcfd91279f/src/docker-in-docker/devcontainer-feature.json#L60

Mounts may also require special consideration. 🤔

docs/docker.md
```diff
@@ -0,0 +1,108 @@
...
+usage of the `root` user inside the container, as the Docker daemon will be run
+under a "fake" root user (via `rootlesskit`). The user inside the workspace can
+then be a 'regular' user without root permissions.
+
+> Note: Once again, we use a custom entrypoint via `ENVBUILDER_INIT_SCRIPT` to
+> start the Docker daemon via `rootlesskit`.
+
+Example:
+
+```
```
mafredri, line 74: Nit: Consistency (alt. all could be `console` too).
```suggestion
```shell
```

examples/docker/02_dind/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM ubuntu:latest
```
mafredri, line 1:
```suggestion
FROM ubuntu:noble
```
Noble is a LTS release, and I think it's good to avoid sudden breakage because of sudden change in Ubuntu release.

examples/docker/03_dind_rootless/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+## Install prerequisites
+RUN apt-get update && \
+  apt-get install -y apt-transport-https curl iproute2 uidmap
+# Install Docker daemon
+RUN curl -fsSL https://get.docker.com/ | sh -s -
+# Add the ubuntu user to the Docker group
+RUN usermod -aG docker ubuntu
+# Add our custom entrypoint
+ADD entrypoint.sh /entrypoint.sh
+RUN chmod +x /entrypoint.sh
```
mafredri, line 11: Git supports executable bit, so we could skip this. We also don't have it in DinD.

examples/docker/03_dind_rootless/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+# Create the XDG_RUNTIME_DIR for our user and set DOCKER_HOST
+ENV XDG_RUNTIME_DIR /run/user/1000
+RUN mkdir -p ${XDG_RUNTIME_DIR} && chown ubuntu:ubuntu ${XDG_RUNTIME_DIR}
+ENV DOCKER_HOST unix:///${XDG_RUNTIME_DIR}/docker.sock
+# Setup rootless mode as the ubuntu user.
+USER ubuntu
+RUN dockerd-rootless-setuptool.sh install
+# Switch the ubuntu user to use the rootless docker context
+RUN docker context use rootless
+RUN mkdir -p /home/ubuntu/.local/share/docker
```
mafredri, line 21: Not necessary in this PR, but future consideration: Should we use best-practices in this Dockerfile, like combining RUN steps (fewer layers) and optimizing for files that change and not. E.g. `COPY entrypoint.sh` is likely to change, whereas `dockerd-rootless-setuptool.sh install` is not, which would suggest we do COPY at the end.

mafredri (approved):
Awesome!

docs/docker.md
johnstcn: 🤦

docs/docker.md
johnstcn: Yeah, the custom entrypoint is still required here. 👍

johnstcn: I added a separate example `03_dind_feature` for this. Works fine with volume mounts as far as I tested, too.

examples/docker/03_dind_rootless/Dockerfile
johnstcn: I went ahead and tidied it up a bit.

Issue #178, opened 2024-05-05
Update kaniko fork


matifali:
Coder's [kaniko](https://github.com/coder/kaniko) fork is outdated and prevents updating some of the dependencies that have vulnerabilities. 
Details can be seen here: https://github.com/coder/envbuilder/security/dependabot

This is a must-do before we launch 1.0
cc: @bpmct

kylecarbs:
https://github.com/coder/kaniko/pull/7

matifali:
Should we add a scheduled workflow to keep it in sync weekly or monthly?
https://github.com/wei/pull looks promising

johnstcn:
Kaniko has since updated to go1.22 which means we also need to do so. Unfortunately our codersdk deps prevent this right now. For now I'm going to manually vendor these in.

mtojek:
@johnstcn Is there anything left here or can we resolve this issue?

johnstcn:
@mtojek https://github.com/coder/kaniko/pull/7 is still un-merged. Some of the tests are failing partially due to some changes we have that upstream does not.

Then https://github.com/coder/envbuilder/pull/195 can be updated and approved.

johnstcn:
This is done now.

PR #194, opened 2024-05-17
chore: remove codersdk dependency

johnstcn:
Part of https://github.com/coder/envbuilder/issues/178

In order to update our branch of Kaniko, we need to first update to go1.22.
This is not currently possible while depending on codersdk.

- Manually vendored relevant parts of `codersdk` and `agentsdk` into `internal/notcodersdk`
- Replaced existing usage of `codersdk` / `agentsdk` with `internal/notcodersdk`
- Added test for coder log sending functionality

Before:

```
$ du -sh ./scripts/envbuilder-amd64 
64M     ./scripts/envbuilder-amd64
```

After:
```
$ du -sh ./scripts/envbuilder-amd64 
45M     ./scripts/envbuilder-amd64
```


dannykopping (approved):
LGTM

Do we anticipate we will be in this state for some time? If so, a `README` in the `notcodersdk` would be helpful for archaeological purposes. _Optionally_ one step further would be to add a script to clone the repo and lint to ensure we aren't drifting too far from the original repo, but this may be overkill for now.

cmd/envbuilder/main.go
```diff
@@ -25,12 +24,21 @@ import (
...
 func main() {
+	cmd := envbuilderCmd()
+	err := cmd.Invoke().WithOS().Run()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "error: %v", err)
+		os.Exit(1)
+	}
+}
+
+func envbuilderCmd() *serpent.Command {
```
mafredri, line 35: Nit: Could probably skip the ptr here. 🤔

Also: Good choice splitting this up.

mafredri (approved):
> notcodersdk

This sounds sneaky. 😂

Issue #70, opened 2023-12-17
workspace file permissions incorrect from COPY instruction


coryb:
Files being added to the workspace via `COPY` and `ADD` instructions in the dockerfile will end up with the original permissions from the workspace mount, not the correct default `root:root` file ownership. 

Note this is only when used with docker on Linux.  

Here is a simple reproduction:
```
$ cat .devcontainer/devcontainer.json
{"build":{"dockerfile":"Dockerfile"}}
```
```
$ cat .devcontainer/Dockerfile
FROM ubuntu
COPY files /files
RUN find /files -ls
```
```
$ docker run --rm -e WORKSPACE_FOLDER=/workspace -e INIT_SCRIPT=/bin/true -v $(pwd):/workspace ghcr.io/coder/envbuilder
...
#2: Running: [/bin/sh -c find /files -ls]
 18395960      4 drwxr-xr-x   2 1000     1000         4096 Dec 17 01:18 /files
 18395961      0 -rw-r--r--   1 1000     1000            0 Dec 17 01:17 /files/some.txt
```

From the Dockerfile spec:
> All new files and directories are created with a UID and GID of 0.

Likely caused by bug in Kaniko https://github.com/GoogleContainerTools/kaniko/issues/2850

The current hacky work-around is to force the --chown:
```
COPY --chown=0:0 files /files
```
Which seems to do the "right" thing:
```
#2: Running: [/bin/sh -c find /files -ls]
 18395960      4 drwxr-xr-x   2 root     root         4096 Dec 17 01:23 /files
 18395961      0 -rw-r--r--   1 root     root            0 Dec 17 01:23 /files/some.txt
```

coryb:
Looks like a PR was created 7mo ago in Kankio, but it is still unmerged https://github.com/GoogleContainerTools/kaniko/pull/2494

Maybe we can cherry pick this commit into the coder/kaniko fork?

johnstcn:
Unfortunately, it looks like the upstream fix breaks user/group ownership in multi-stage builds.

PR #202, opened 2024-05-20
fix: revert setting default UserID and GroupID to 0:0

johnstcn:
After looking into the failing unit tests some more, it looks like https://github.com/coder/kaniko/commit/9f83bc8595cde0965de0b5a17e6770e5c2b1321f actually does cause divergence from Docker behaviour.

I've included an integration test adapted from GoogleContainerTools/kaniko that reproduces the issue. It affects file permissions when changing permissions in a multi-stage build but does not appear to affect single-stage builds.

This test fails with the commit, and succeeds without. The fix has not been merged upstream yet as it would break legacy users. For now, we will revert to the upstream behaviour.

This will unfortunately re-break https://github.com/coder/envbuilder/issues/70 but has a simple workaround of using `COPY --chown 0:0`.

### DIFF
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -94,7 +94,7 @@ func TestInitScriptInitCommand(t *testing.T) {
 }
 
 func TestUidGid(t *testing.T) {
-
+	t.Parallel()
 	t.Run("MultiStage", func(t *testing.T) {
 		t.Parallel()
 
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -93,6 +93,74 @@ func TestInitScriptInitCommand(t *testing.T) {
 	require.NoError(t, ctx.Err(), "init script did not execute for legacy env vars")
 }
 
+func TestUidGid(t *testing.T) {
+
+	t.Run("MultiStage", func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


integration/integration_test.go
```diff
@@ -93,6 +93,74 @@ func TestInitScriptInitCommand(t *testing.T) {
...
+			envbuilderEnv("GIT_URL", srv.URL),
+			envbuilderEnv("DOCKERFILE_PATH", "Dockerfile"),
+		}})
+		require.NoError(t, err)
+	})
+
+	t.Run("SingleStage", func(t *testing.T) {
+		t.Parallel()
+
+		dockerFile := fmt.Sprintf(`FROM %s AS builder
```
johnstcn, line 135:
```suggestion
		dockerFile := fmt.Sprintf(`FROM %s
```

johnstcn:
> What's the divergence? As far as Docker spec is concerned, root is the default:

In a multi-stage build, the UIDs/GIDs of the source appear to be retained.

mafredri (approved):
I think breaking multi stage builds is a good enough reason to revert this for the time being. Let's revisit later if we can bring this in line with Docker behavior.

PR #222, opened 2024-06-05
ci: build and push image for `main` branch

matifali:
This pull request updates the `ci` workflow to build and push a Docker image for the `main` branch. 

Image is pushed to `ghcr.io/coder/envbuilder-preview` to not pollute the manually tagged versions at `ghcr.io/coder/envbuilder`

This will help test the features that are under development. 

Each image is tagged as `latest` and `$(./scripts/version.sh)-dev-$(git rev-parse --short HEAD)` e.g.,
1. `ghcr.io/coder/envbuilder-preview:latest`
2. `ghcr.io/coder/envbuilder-preview:v0.2.9-dev-ab86184`

---
Alternatively, we can push `ghcr.io/coder/envbuilder:main` to the same package.


.github/workflows/ci.yaml
```diff
@@ -67,3 +68,49 @@ jobs:
...
+      - name: Build
+        if: github.event_name == 'pull_request'
+        run: |
+          VERSION=$(./scripts/version.sh)-dev-$(git rev-parse --short HEAD)
+          BASE=ghcr.io/coder/envbuilder-preview
+
+          ./scripts/build.sh \
+            --arch=amd64 \
+            --base=$BASE \
+            --tag=$VERSION
```
matifali, line 102: I am not building a multi-arch here, as this step only ensures that Dockerfile builds fine.

johnstcn: we should probably still build multi-arch as some changes may only cause issues with certain architectures

dannykopping (approved):
LGTM

To what extent has this workflow been tested?
Has the workflow been linted?

scripts/build.sh
```diff
@@ -63,7 +62,7 @@ else
   args+=( --load )
 fi
 
-docker buildx build "${args[@]}" -t $base:$tag -t $base:latest -f Dockerfile .
+docker buildx build --builder $BUILDER_NAME "${args[@]}" -t $base:$tag -t $base:latest -f Dockerfile .
```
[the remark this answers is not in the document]
matifali: Yes though about this but will require hacking the build script to work differently for tagged releases. Currently the script has a hardcoded `$base:latest` so I opted for changing the `$base`.

.github/workflows/ci.yaml
```diff
@@ -67,3 +68,49 @@ jobs:
...
       - name: Check format
         run: ./scripts/check_fmt.sh
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          # Needed to get older tags
+          fetch-depth: 0
```
johnstcn, line 78: we probably don't need the tags if we're just building a preview image on `main`?

scripts/build.sh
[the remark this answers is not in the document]
johnstcn: fair, not blocking

johnstcn (approved):

matifali:
@dannykopping I have tested the workflow except the push image step, but will pass this through the linter. Thanks for calling it.

.github/workflows/ci.yaml
matifali: Ok. Our build script uses `--load` when `--push` is not specified, which cannot be used for multi arch images. 
https://github.com/docker/buildx/issues/59

I didn't look deep if we actually needed to load the built images during local development otherwise I can just remove `--load` and build a multi arch image. 

While calling the script locally we only build with `$current` arch so `--load` works fine.

johnstcn:
@matifali is this good to merge?

matifali:
Yes. Once the CI passes it should be good to merge.

PR #233, opened 2024-06-12
chore: integration: add test for pushing to cache repo that requires auth

johnstcn:
Adds a gap in our testing for the case when the cache repo requires auth.
NOTE: this shows that being unable to push the image will fail the build outright.
Do we maybe want to relax this?

### DIFF
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -24,6 +24,7 @@ import (
 	"github.com/coder/envbuilder"
 	"github.com/coder/envbuilder/devcontainer/features"
 	"github.com/coder/envbuilder/testutil/gittest"
+	"github.com/coder/envbuilder/testutil/mwtest"
 	"github.com/coder/envbuilder/testutil/registrytest"
 	clitypes "github.com/docker/cli/cli/config/types"
 	"github.com/docker/docker/api/types"
@@ -1362,7 +1363,7 @@ func setupInMemoryRegistry(t *testing.T, opts setupInMemoryRegistryOpts) string 
 	t.Helper()
 	tempDir := t.TempDir()
 	regHandler := registry.New(registry.WithBlobHandler(registry.NewDiskBlobHandler(tempDir)))
-	authHandler := gittest.BasicAuthMW(opts.Username, opts.Password)(regHandler)
+	authHandler := mwtest.BasicAuthMW(opts.Username, opts.Password)(regHandler)
 	regSrv := httptest.NewServer(authHandler)
 	t.Cleanup(func() { regSrv.Close() })
 	regSrvURL, err := url.Parse(regSrv.URL)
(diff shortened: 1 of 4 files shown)


integration/integration_test.go
```diff
@@ -1101,6 +1107,130 @@ func TestPushImage(t *testing.T) {
 		require.NoError(t, err)
 	})
 
+	t.Run("CacheAndPushAuth", func(t *testing.T) {
```
johnstcn, line 1110: review: I may look into making this more of a table-driven test, as it's getting pretty unwieldy

testutil/gittest/gittest.go
```diff
@@ -249,18 +249,3 @@ func WriteFile(t *testing.T, fs billy.Filesystem, path, content string) {
 	err = file.Close()
 	require.NoError(t, err)
 }
-
```
johnstcn, line 252: review: moved as this isn't specific to testing git stuff

integration/integration_test.go
dannykopping: I was gonna suggest the same.

integration/integration_test.go
```diff
@@ -1101,6 +1107,130 @@ func TestPushImage(t *testing.T) {
...
+		require.NoError(t, err, "expected image to be present after build + push")
+
+		// Then: re-running envbuilder with GET_CACHED_IMAGE should succeed
+		_, err = runEnvbuilder(t, options{env: []string{
+			envbuilderEnv("GIT_URL", srv.URL),
+			envbuilderEnv("CACHE_REPO", testRepo),
+			envbuilderEnv("GET_CACHED_IMAGE", "1"),
+			envbuilderEnv("DOCKER_CONFIG_BASE64", base64.StdEncoding.EncodeToString(regAuthJSON)),
+		}})
+		require.NoError(t, err)
```
dannykopping, line 1178: Is there any more specificity we could add here? The absence of an error is fine, but validating the cached image exists explicitly would be better.

dannykopping (approved):
LGTM

> NOTE: this shows that being unable to push the image will fail the build outright.
Do we maybe want to relax this?

The pushed image is just an optimization to speed up the subsequent runs, correct?
It feels like _Principle of Least Surprise_ here would be to complete the primary objective (which is building), and any ancillary steps to optimize should not fail that.

johnstcn:
> The pushed image is just an optimization to speed up the subsequent runs, correct? It feels like _Principle of Least Surprise_ here would be to complete the primary objective (which is building), and any ancillary steps to optimize should not fail that.

As a counterpoint, I can foresee this being used in a CI scenario along with something like `ENVBUILDER_INIT_SCRIPT=exit 0` in order to 'pre-warm' the cache repo. I'm not sure which behaviour is most desirable here.

integration/integration_test.go
johnstcn: see L1168/1169

mtojek (approved):
Good idea 👍

integration/integration_test.go
johnstcn: I had a try and there's a bit too much difference behaviour between the various test cases to really make it worthwhile IMO compared with the cognitive overhead of DRYing this.

Issue #216, opened 2024-06-01
Embed the envbuilder binary in images


mafredri:
Depends on https://github.com/coder/envbuilder/pull/197

As part of #128, we want to embed the envbuilder binary in container images that are pushed to a registry (see #213 for the push implementation).

This could be as simple as appending `COPY /.envbuilder/bin/envbuilder /.envbuilder/bin/envbuilder` to the Dockerfile. The binary could be located elsewhere too, so a lookup may be necessary.

This will embed an eventually stale envbuilder binary in the image, but we consider this OK as our recommendation is to lock the envbuilder version, updating it will result in a new build.

**Acceptance Criteria:**

1. Build an image using envbuilder with `ENVBUILDER_CACHE_REPO=localhost:5000/envbuilder-cache` or similar (after #197 is merged)
2. Pull the resulting image
3. Validate that the same `envbuilder` binary is present in the same location in the resulting image

BrunoQuaresma:
I think it is done already. When I try to run envbuilder inside of a container from an integration test it works so looks like the binary is already in the image.

```golang
execContainer(t, ctr, "/.envbuilder/bin/envbuilder --help")
```

```
USAGE:
  envbuilder

OPTIONS:
      --base-image-cache-dir string, $ENVBUILDER_BASE_IMAGE_CACHE_DIR
          The path to a directory where the base image can be found. This should
          be a read-only directory solely mounted for the purpose of caching the
          base image.

      --build-context-path string, $ENVBUILDER_BUILD_CONTEXT_PATH
          Can be specified when a DockerfilePath is specified outside the base
          WorkspaceFolder. This path MUST be relative to the WorkspaceFolder
          path into which the repo is cloned.

      --cache-repo string, $ENVBUILDER_CACHE_REPO
          The name of the container registry to push the cache image to. If this
          is empty, the cache will not be pushed.

      --cache-ttl-days int, $ENVBUILDER_CACHE_TTL_DAYS
          The number of days to use cached layers before expiring them. Defaults
          to 7 days.

      --coder-agent-subsystem string-array, $CODER_AGENT_SUBSYSTEM
          Coder agent subsystems to report when forwarding logs...+6238 more
```

johnstcn:
^ It needs to be in the built and pushed image. See the acceptance criteria I added in the issue description.

BrunoQuaresma:
Looks like it is already there. Here is the testing that I'm using to check that.

```golang
[...]

PR #234, opened 2024-06-12
feat: embed binary in image when pushing image

BrunoQuaresma:
Related to https://github.com/coder/envbuilder/issues/216

Embeds the envbuilder binary in the image under the same path when pushing.

Depends on https://github.com/coder/kaniko/pull/16


integration/integration_test.go
```diff
@@ -1354,6 +1354,45 @@ COPY --from=a /root/date.txt /date.txt`, testImageAlpine, testImageAlpine),
...
+	}})
+	require.NoError(t, err)
+
+	_, err = remote.Image(ref)
+	require.NoError(t, err, "expected image to be present after build + push")
+
+	ctr, err := runEnvbuilder(t, options{env: []string{
+		envbuilderEnv("FALLBACK_IMAGE", ref.String()),
+	}})
+	require.NoError(t, err)
```
johnstcn, line 1390: What we really want to do is the equivalent of 

```
docker run -it --rm localhost:5000/test-embed-binary-image:latest /bin/sh -c '[[ -f /.envbuilder/bin/envbuilder ]] && echo exists'
```

What you have here actually is more similar to:

```
docker run -it --rm -e FALLBACK_IMAGE=localhost:5000/test-embed-binary-image:latest /bin/sh -c '[[ -f /.envbuilder/bin/envbuilder ]] && echo exists'
```

The `envbuilder` binary that this is testing is from having run envbuilder itself, and not from the image that was pushed to the local registry.

integration/integration_test.go
```diff
@@ -1384,13 +1385,42 @@ func TestEmbedBinaryImage(t *testing.T) {
...
+	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
+	require.NoError(t, err)
+	t.Cleanup(func() {
+		cli.Close()
+	})
+
+	// Pull the image we just built
+	rc, err := cli.ImagePull(ctx, ref.String(), image.PullOptions{})
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = rc.Close() })
```
BrunoQuaresma, line 1398: @johnstcn why not use `defer rc.Close()` instead?

johnstcn: I think that would also work here, I just prefer `t.Cleanup` inside tests.

integration/integration_test.go
```diff
@@ -1354,6 +1355,74 @@ COPY --from=a /root/date.txt /date.txt`, testImageAlpine, testImageAlpine),
 	})
 }
 
+func TestEmbedBinaryImage(t *testing.T) {
```
mtojek, line 1358: I would annotate this test file with given/when/then markers.

integration/integration_test.go
```diff
@@ -1354,6 +1355,74 @@ COPY --from=a /root/date.txt /date.txt`, testImageAlpine, testImageAlpine),
...
+		envbuilderEnv("GIT_URL", srv.URL),
+		envbuilderEnv("CACHE_REPO", testRepo),
+		envbuilderEnv("PUSH_IMAGE", "1"),
+	}})
+	require.NoError(t, err)
+
+	_, err = remote.Image(ref)
+	require.NoError(t, err, "expected image to be present after build + push")
+
+	ctx := context.Background()
```
mtojek, line 1388: It is not a problem at all, but a cleaner approach would be:

```
ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
defer cancel()
```

envbuilder.go
```diff
@@ -521,7 +526,7 @@ func Run(ctx context.Context, options Options) error {
 				// https://github.com/coder/envbuilder/pull/114
 				RegistryMirrors: registryMirror,
 			},
-			SrcContext: buildParams.BuildContext,
```
mtojek, line 524: Is this relevant to this PR?

envbuilder.go
```diff
@@ -342,6 +343,21 @@ func Run(ctx context.Context, options Options) error {
...
+	if options.PushImage {
+		// Copy the envbuilder binary into the build context.
+		buildParams.DockerfileContent = buildParams.DockerfileContent + "\n" +
+			fmt.Sprintf("COPY %s %s", ".envbuilder", "/.envbuilder")
+
+		log.Println("FIXME show me the Dockerfile now: " + buildParams.DockerfileContent)
+
+		binPath := filepath.Join(MagicDir, "bin", "envbuilder")
+		dst := filepath.Join(buildParams.BuildContext, binPath)
+		err := copyFile(binPath, dst)
```
mtojek, line 355: @BrunoQuaresma @johnstcn 

Alright, I spent some time digging into this. I don't get one thing about the Docker lifecycle: 

```
    integration_test.go:1633: "#2: COPY .envbuilder /.envbuilder"
    integration_test.go:1633: "Failed to build: do build: error building stage: failed to execute command: copying dir: creating file: open /.envbuilder/bin/envbuilder: text file busy"
    integration_test.go:1633: "Falling back to the default image..."
    integration_test.go:1633: "error: do build: error building stage: failed to execute command: copying dir: creating file: open /.envbuilder/bin/envbuilder: text file busy: no fallback image has been specified"
    integration_test.go:1383:
        	Error Trace:	/code/envbuilder/integration/integration_test.go:1383
        	Error:      	Received unexpected error:
        	            	error: do build: error building stage: failed to execute command: copying dir: creating file: open /.envbuilder/bin/envbuilder: text file busy: no fallback image has been specified
        	Test:       	TestEmbedBinaryImage
--- FAIL: TestEmbedBinaryImage (0.62s)
```

As you see `COPY` failed because the binary, which is already there, has been started? And the binary is there because in the integration test we use `localhost:5000/envbuilder-test-alpine:latest`. I see we have 2 options:
1. Do not use Docker image with `envbuilder` as base.
2. Check if binary is there and do not copy it.

Side question: why is the binary started? Is it always like that with kaniko?

johnstcn: > Do not use Docker image with envbuilder as base.

The image `localhost:5000/envbuilder-test-alpine:latest` should not contain the `envbuilder` binary. It simply a locally-cached version of `alpine:latest`.

```
docker run -it --rm localhost:5000/envbuilder-test-alpine:latest ls -la /.envbuilder
ls: /.envbuilder: No such file or directory
```

> Side question: why is the binary started?

The `envbuilder` binary is going to be running inside the container no matter what, as that's the container entrypoint.

mtojek: OK, so it seems that `COPY ` operation is not necessary? See the build fails due to `text file busy`. This is the part I'm trying to understand.

johnstcn: OK, instead of copying the binary manually, we need to instead tell Kaniko to not skip over `/.envbuilder/bin/envbuilder`.

BrunoQuaresma: @johnstcn just double checking, are you going to take over this PR?

johnstcn: yep

mtojek (approved):
👍 

It is really fortunate that API was exposed.

envbuilder.go
```diff
@@ -404,9 +404,29 @@ func Run(ctx context.Context, options Options) error {
 		util.AddToDefaultIgnoreList(util.IgnoreListEntry{
 			Path:            ignorePath,
 			PrefixMatchOnly: false,
+			AllowedPaths:    map[string]struct{}{},
```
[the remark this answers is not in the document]
johnstcn: Now that I think of it, all of that is just to avoid a panic on assigning to a nil map. However, using a `[]string` instead doesn't have that issue, and might even be a tiny bit faster.

mtojek (approved):

Issue #217, opened 2024-06-01
Sanitize the final container image for envbuilder


mafredri:
As part of https://github.com/coder/envbuilder/issues/128, we want to sanitize the container images that are pushed to a registry (see https://github.com/coder/envbuilder/pull/213 for the push implementation).

The motivation is that when running a pre-built image, `envbuilder` should be able to perform the tasks listen in `devcontainer.json` to finalize the environment (e.g. run scripts).

Example sanitization directives that can be appended to the Dockerfile:

```dockerfile
USER root
WORKDIR /
ENTRYPOINT ["/.envbuilder/bin/envbuilder"]
```

Note, `ENTRYPOINT` is dependent on #216.

PR #235, opened 2024-06-12
chore: make sure directives are correct

BrunoQuaresma:
Closes https://github.com/coder/envbuilder/issues/217

### DIFF
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -342,6 +342,12 @@ func Run(ctx context.Context, options Options) error {
 		}
 	}
 
+	// Make sure the Dockerfile is using the correct directives to run envbuilder
+	buildParams.DockerfileContent = buildParams.DockerfileContent + "\n" +
+		"USER root \n" +
+		"WORKDIR / \n" +
+		"ENTRYPOINT [\"/.envbuilder/bin/envbuilder\"]"
+
 	HijackLogrus(func(entry *logrus.Entry) {
 		for _, line := range strings.Split(entry.Message, "\r") {
 			options.Logger(notcodersdk.LogLevelInfo, "#%d: %s", stageNumber, color.HiBlackString(line))
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -342,11 +342,13 @@ func Run(ctx context.Context, options Options) error {
 		}
 	}
 
-	// Make sure the Dockerfile is using the correct directives to run envbuilder
-	buildParams.DockerfileContent = buildParams.DockerfileContent + "\n" +
-		"USER root \n" +
-		"WORKDIR / \n" +
-		"ENTRYPOINT [\"/.envbuilder/bin/envbuilder\"]"
+	if options.PushImage {
+		// Make sure the Dockerfile is using the correct directives to run envbuilder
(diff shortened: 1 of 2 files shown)


johnstcn (commented):
I verified the behaviour is correct 👍 
Do we need to modify these directives if we are not pushing the image?

BrunoQuaresma:
@johnstcn Good question, from what I can see on https://github.com/coder/envbuilder/issues/217 we should not. I'm going to submit a fix.

integration/integration_test.go
```diff
@@ -1354,6 +1354,48 @@ COPY --from=a /root/date.txt /date.txt`, testImageAlpine, testImageAlpine),
...
+	require.NoError(t, err)
+
+	image, err := remote.Image(ref)
+	require.NoError(t, err, "expected image to be present after build + push")
+	configFile, err := image.ConfigFile()
+	require.NoError(t, err, "expected image to return a config file")
+
+	require.Equal(t, configFile.Config.User, "root", "value does not match any of the possible root user values.")
+	require.Equal(t, configFile.Config.WorkingDir, "/", "expected image to have root working directory")
+	require.Equal(t, configFile.Config.Entrypoint, []string{"/.envbuilder/bin/envbuilder"}, "expected image to have envbuilder entrypoint")
```
johnstcn, line 1396: probably a good idea to assert the opposite in some other test when we don't push

envbuilder.go
```diff
@@ -342,6 +342,14 @@ func Run(ctx context.Context, options Options) error {
 		}
 	}
 
+	if options.PushImage {
+		// Make sure the Dockerfile is using the correct directives to run envbuilder
+		buildParams.DockerfileContent = buildParams.DockerfileContent + "\n" +
```
mtojek, line 347: I guess that you'll have a merge conflict with the other PR: https://github.com/coder/envbuilder/issues/216

integration/integration_test.go
```diff
@@ -1354,6 +1354,48 @@ COPY --from=a /root/date.txt /date.txt`, testImageAlpine, testImageAlpine),
...
+	t.Parallel()
+
+	srv := createGitServer(t, gitServerOptions{
+		files: map[string]string{
+			".devcontainer/entrypoint.sh": "#!/bin/sh\necho 'Hello, world!'",
+			".devcontainer/Dockerfile": "FROM " + testImageAlpine + "\n" +
+				"RUN addgroup -S coder && adduser -S coder -G coder \n" +
+				"USER coder\n" +
+				"WORKDIR /app\n" +
+				"ENTRYPOINT [\"./entrypoint.sh\"]",
```
mtojek, line 1367: maybe add a comment here that these settings should be overridden by envbuilder

integration/integration_test.go
BrunoQuaresma: If they are not pushed, what reference can we use to get the local image and inspect it?

BrunoQuaresma:
This is blocked by https://github.com/coder/envbuilder/issues/216 because it needs to have the embed binary to set the entrypoint correctly

johnstcn:
closing in favour of https://github.com/coder/envbuilder/pull/246

PR #239, opened 2024-06-16
Update README.md

paviln:
Hello,

It seems like ENVBUILDER_GIT_USERNAME/PASSWORD is not used, and GIT_USERNAME/PASSWORD works.

Can someone explain why that is?

If this is a mistake in docs, it can be updated by this pull request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,11 +146,11 @@ Two methods of authentication are supported:
 ### HTTP Authentication
 
 If `ENVBUILDER_GIT_URL` starts with `http://` or `https://`, envbuilder will
-authenticate with `ENVBUILDER_GIT_USERNAME` and `ENVBUILDER_GIT_PASSWORD`, if set.
+authenticate with `GIT_USERNAME` and `GIT_PASSWORD`, if set.
 
 For access token-based authentication, follow the following schema (if empty, there's no need to provide the field):
 
-| Provider     | `ENVBUILDER_GIT_USERNAME` | `ENVBUILDER_GIT_PASSWORD` |
+| Provider     | `GIT_USERNAME` | `GIT_PASSWORD` |
 | ------------ | ------------------------- | ------------------------- |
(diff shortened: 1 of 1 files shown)


matifali:
@paviln Hi, you are right about the difference in variable names. `envbuilder`'s `main` branch uses the new `ENVBUILDER_` names, but it has not been published yet, and the last published version [`v0.2.9`](https://github.com/coder/envbuilder/pkgs/container/envbuilder/193424369?tag=0.2.9) does not support it.

It should be fixed when a new version of `envbuilder` is published.

ggjulio:
I was fooled by the readme too.

There is other files to update as well 
- https://github.com/coder/envbuilder/blob/8c5c50cad3ae396b73619d36da43ef2ff0ac0be8/testdata/options.golden#L94
- https://github.com/coder/envbuilder/blob/8c5c50cad3ae396b73619d36da43ef2ff0ac0be8/docs/docker.md?plain=1#L23

README.md
```diff
@@ -146,11 +146,11 @@ Two methods of authentication are supported:
 ### HTTP Authentication
 
 If `ENVBUILDER_GIT_URL` starts with `http://` or `https://`, envbuilder will
```
ggjulio, line 148: It's `GIT_URL` now right ? 

```diff
- If `ENVBUILDER_GIT_URL` starts with `http://` or `https://`, envbuilder will
+ If `GIT_URL` starts with `http://` or `https://`, envbuilder will
```

BrunoQuaresma:
Hey folks, the main version of envbuilder uses the `ENVBUILDER` prefix for the env variables so all the README is using it and we plan to keep it. For now, you can use the README from the latest published version here https://github.com/coder/envbuilder/tree/v0.2.9.

johnstcn:
See also https://github.com/coder/envbuilder/releases/tag/v1.0.0-rc.0

PR #243, opened 2024-06-17
Fix FeatureContexts keys

aaronlehmann:
This needs to be updated to correspond to the changes in #205.


devcontainer/devcontainer_test.go
```diff
@@ -108,6 +109,38 @@ WORKDIR `+featureTwoDir+`
 ENV POTATO=example
 RUN VERSION="potato" _CONTAINER_USER="1000" _REMOTE_USER="1000" ./install.sh
 USER 1000`, params.DockerfileContent)
+	})
+
+	t.Run("WithBuildContexts", func(t *testing.T) {
+		params, err := dc.Compile(fs, "", magicDir, "", "", true, os.LookupEnv)
```
johnstcn, line 115: nit, non-blocking: I know this is a pre-existing convention, but using `os.LookupEnv` here has potential to introduce test flakes. What would you think about swapping it out with a simple stub `func()` instead?

johnstcn (approved):
Thanks for the contribution!

devcontainer/devcontainer_test.go
aaronlehmann: Good idea - done.

johnstcn (approved):

PR #247, opened 2024-06-24
Use a deterministic target path for feature dir mounts with useBuildContexts

aaronlehmann:
This avoids caching issues when using the Dockerfile produced in `useBuildContexts` mode. See
https://github.com/coder/envbuilder/pull/205#issuecomment-2181962174 for context.

### DIFF
--- a/devcontainer/devcontainer_test.go
+++ b/devcontainer/devcontainer_test.go
@@ -131,13 +131,13 @@ FROM localhost:5000/envbuilder-test-codercom-code-server:latest
 
 USER root
 # Rust tomato - Example description!
-WORKDIR `+featureOneDir+`
+WORKDIR /.envbuilder/features/one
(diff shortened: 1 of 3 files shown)


kylecarbs (commented):
Should we join it with the `MagicDir` instead? That way we don't create two root-level ones.

aaronlehmann:
Works for me! I've changed the prefix to `/.envbuilder/features`. I went ahead and hardcoded this because this package won't be able to import the root `envbuilder` package, but let me know if you prefer to pass in `MagicDir` instead.

kylecarbs:
I think duplicating it is fine!

kylecarbs (approved):

kylecarbs:

aaronlehmann:
Thanks!

PR #248, opened 2024-06-25
fix(devcontainer): correctly parse feature with digest

maxbrunet:
Hello! 👋

Parsing an image with a digest like `ghcr.io/coder/envbuilder:0.2.9@sha256:b51153d8cc1235986707d5f61949884444f2b4c4f64b15b5506cb5484bfa8a93` using `NewTag` does not work:

```
repository can only contain the characters `abcdefghijklmnopqrstuvwxyz0123456789_-./`: coder/envbuilder:0.2.9@sha256
```

`ParseReference` supports both tag and digest, and produces the same result.


johnstcn (approved):
Thanks for the contribution! `name.ParseReference` uses `NewTag` under the hood (falling back to `NewDigest`), so this looks fine to me. 👍

PR #255, opened 2024-06-28
fix(remount): relocate libraries along with their symlinks

### LINKED ISSUE #143: Investigate GPU support
Some users will want to mount a GPU to an envbuilder-backed workspace. Can we investigate in which scenarios (if any) this works today and if/how we can patch upstream Kaniko to improve the experience?

## Related

- Kaniko issue: https://github.com/GoogleContainerTools/kaniko/issues/3006
- [Internal Slack message](https://codercom.slack.com/archives/C015YDF37AT/p1714057064019159)

maxbrunet:
This PR adds:

1. Look up the library directory, the value differs for Debian-based distros (the initial envbuilder image is non-Debian, but the final image may be Debian)
2. Find the symlinks pointing to mounts in the library directory
3. Temporarily move the library symlinks pointing to mounts to the magic directory while relocating the mounts
4. Look up the new library directory (in case it has changed)
5. Move back the library symlinks and mounts to the new library directory

After that the container should behave like a regular container created by the NVIDIA container runtime. Of course/unfortunately, the process of mounting/unmounting requires GPU containers to run with privileges:

> Appropriate privilege (Linux: the `CAP_SYS_ADMIN` capability) is required to mount/umount filesystems.

https://www.man7.org/linux/man-pages/man2/mount.2.html
https://www.man7.org/linux/man-pages/man2/umount.2.html

The logic is not generalized to any symlinks or any directories, it only aims at providing compatibility with the NVIDIA container runtime for now.

More context can be found in this comment https://github.com/coder/envbuilder/issues/143#issuecomment-2192405828

Tested with the following images:

* `docker.io/library/debian:bookworm`
* `docker.io/library/fedora:40`
* `nvcr.io/nvidia/pytorch:24.05-py3`

Closes #143

### DIFF
--- a/internal/ebutil/libs.go
+++ b/internal/ebutil/libs.go
@@ -1,0 +1,78 @@
+package ebutil
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+// Based on https://github.com/NVIDIA/libnvidia-container/blob/v1.15.0/src/common.h#L29
+
+const usrLibDir = "/usr/lib64"
+
+const debianVersionFile = "/etc/debian_version"
+
+// getLibDir returns the library directory. It returns a multiarch directory if
+// the distribution is Debian or a derivative.
+//
+// Based on https://github.com/NVIDIA/libnvidia-container/blob/v1.15.0/src/nvc_container.c#L152-L165
+func getLibDir(m mounter) (string, error) {
+	// Debian and its derivatives use a multiarch directory scheme.
+	if _, err := m.Stat(debianVersionFile); err != nil && !errors.Is(err, os.ErrNotExist) {
+		return "", fmt.Errorf("check if debian: %w", err)
+	} else if err == nil {
+		return usrLibMultiarchDir, nil
+	}
+
+	return usrLibDir, nil
+}
+
+// getLibsSymlinks returns the stats for all library symlinks if the library
+// directory exists.
+func getLibsSymlinks(m mounter, libDir string) (map[string][]string, error) {
+	des, err := m.ReadDir(libDir)
+	if err != nil {
+		return nil, fmt.Errorf("read directory %s: %w", libDir, err)
+	}
+
+	libsSymlinks := make(map[string][]string)
+	for _, de := range des {
+		if de.IsDir() {
+			continue
+		}
+
+		if de.Type()&os.ModeSymlink != os.ModeSymlink {
+			// Not a symlink. Skip.
+			continue
+		}
+
+		symlink := filepath.Join(libDir, de.Name())
+		path, err := m.EvalSymlinks(symlink)
+		if err != nil {
+			return nil, fmt.Errorf("eval symlink %s: %w", symlink, err)
+		}
+
+		path = filepath.Base(path)
+
+		if _, ok := libsSymlinks[path]; !ok {
+			libsSymlinks[path] = make([]string, 0, 1)
+		}
+
+		libsSymlinks[path] = append(libsSymlinks[path], de.Name())
+	}
+
+	return libsSymlinks, nil
+}
+
+// moveLibSymlinks moves a list of symlinks from source to destination directory.
+func moveLibSymlinks(m mounter, symlinks []string, srcDir, destDir string) error {
+	for _, l := range symlinks {
+		oldpath := filepath.Join(srcDir, l)
+		newpath := filepath.Join(destDir, l)
+		if err := m.Rename(oldpath, newpath); err != nil {
+			return fmt.Errorf("move symlink %s => %s: %w", oldpath, newpath, err)
+		}
+	}
+	return nil
+}
--- a/internal/ebutil/libs.go
+++ b/internal/ebutil/libs.go
@@ -7,17 +7,25 @@ import (
 	"path/filepath"
 )
 
-// Based on https://github.com/NVIDIA/libnvidia-container/blob/v1.15.0/src/common.h#L29
+// Container runtimes like NVIDIA mount individual libraries into the container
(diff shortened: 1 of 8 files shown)


internal/ebutil/libs.go
```diff
@@ -0,0 +1,78 @@
...
+		return usrLibMultiarchDir, nil
+	}
+
+	return usrLibDir, nil
+}
+
+// getLibsSymlinks returns the stats for all library symlinks if the library
+// directory exists.
+func getLibsSymlinks(m mounter, libDir string) (map[string][]string, error) {
+	des, err := m.ReadDir(libDir)
```
mtojek, line 34: nit: what does `des` stand for? is it just `fi`, file info?

internal/ebutil/libs.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+
+		symlink := filepath.Join(libDir, de.Name())
+		path, err := m.EvalSymlinks(symlink)
+		if err != nil {
+			return nil, fmt.Errorf("eval symlink %s: %w", symlink, err)
+		}
+
+		path = filepath.Base(path)
+
```
mtojek, line 57: nit: no space needed

internal/ebutil/remount.go
```diff
@@ -87,33 +110,51 @@ outer:
...
 		return fmt.Errorf("ensure path: %w", err)
 	}
+
 	if !stat.IsDir() {
 		f, err := m.OpenFile(dest, os.O_CREATE, 0o640)
 		if err != nil {
 			return fmt.Errorf("ensure file path: %w", err)
 		}
-		defer f.Close()
+		f.Close()
```
mtojek, line 140: nit: please comment why it can't be `defer`red

internal/ebutil/libs_ppc64le.go
```diff
@@ -0,0 +1,7 @@
+//go:build ppc64le
+
+package ebutil
+
+// Based on https://github.com/NVIDIA/libnvidia-container/blob/v1.15.0/src/common.h#L44
```
mtojek, line 5: I'm hesitant to add many architectures if we don't have intention to actively support them. Can we just limit it to arm64 and amd64?

internal/ebutil/remount.go
```diff
@@ -44,15 +45,36 @@ func tempRemount(m mounter, logf func(notcodersdk.LogLevel, string, ...any), bas
 		return func() error { return nil }, fmt.Errorf("get mounts: %w", err)
 	}
 
+	libDir, err := getLibDir(m)
```
mtojek, line 48: nit: `s/getLibDir/libraryDirectoryPath`

internal/ebutil/remount.go
```diff
@@ -44,15 +45,36 @@ func tempRemount(m mounter, logf func(notcodersdk.LogLevel, string, ...any), bas
...
+	if err != nil {
+		return func() error { return nil }, fmt.Errorf("get lib directory: %w", err)
+	}
+
+	libsSymlinks, err := getLibsSymlinks(m, libDir)
```
mtojek, line 53: nit: `s/getLibsSymlinks/libraryDirectorySymlinks`

internal/ebutil/remount.go
```diff
@@ -87,33 +110,51 @@ outer:
...
 	}
+
 	if err := m.Mount(src, dest, "bind", syscall.MS_BIND, ""); err != nil {
 		return fmt.Errorf("bind mount %s => %s: %w", src, dest, err)
 	}
+
 	if err := m.Unmount(src, 0); err != nil {
 		return fmt.Errorf("unmount orig src %s: %w", src, err)
 	}
+
```
mtojek, line 157: nit: no empty line

internal/ebutil/libs.go
```diff
@@ -0,0 +1,78 @@
...
+	} else if err == nil {
+		return usrLibMultiarchDir, nil
+	}
+
+	return usrLibDir, nil
+}
+
+// getLibsSymlinks returns the stats for all library symlinks if the library
+// directory exists.
+func getLibsSymlinks(m mounter, libDir string) (map[string][]string, error) {
```
mtojek, line 33: I'm curious if we need a map here. Can we flatten the structure?

internal/ebutil/libs_ppc64le.go
johnstcn: Seconded; I don't foresee us supporting this arch in the future and the presence of this file may give the wrong impression.

internal/ebutil/libs.go
johnstcn, line 1: I would love to have a big comment here explaining why we need to do this and linking to your fantastic and detailed comment!

https://github.com/coder/envbuilder/issues/143#issuecomment-2192405828

johnstcn (approved):
Thank you for this contribution!

I validated this fix on a Fedora 40 system with the NVidia runtime etc. installed, using both Docker (v27.0.2) and K3s (v1.29.6).

In both cases, I was able to successfully build the image `nvcr.io/nvidia/k8s/cuda-sample:vectoradd-cuda10.2`, run `nvidia-smi` and `/tmp/vectorAdd`.

internal/ebutil/libs.go
maxbrunet: It's `[]os.DirEntry` (short for "directory entries")


https://pkg.go.dev/os#ReadDir

internal/ebutil/libs.go
maxbrunet: The mapping `library` (e.g. `libcuda.so.550.54.15`, what could be a mountpoint) -> `symlinks` (e.g. `[libcuda.so.1, libcuda.so]`). The function comment was outdated, I have updated it.

mtojek (approved):
👍  👍

PR #258, opened 2024-07-02
fix: add usrLibMultiarchDir for 32-bit ARM, whatever that means

johnstcn:
Fixes https://github.com/coder/envbuilder/actions/runs/9757361690


mtojek (approved):
For such PRs we should an option, self-approved :)

PR #259, opened 2024-07-02
fix(remount): correct usrLibMultiarchDir value for 32-bit ARM

maxbrunet:
This constant is for 64-bit systems. 32-bit ARM is not supported.
If ever it becomes supported, it should be handled with a `usrLib32MultiarchDir` constant.

Using `/var/empty` for unsupported directories like NVIDIA: https://github.com/NVIDIA/libnvidia-container/blob/v1.15.0/src/common.h#L53

Follow up to #258

### DIFF
--- a/internal/ebutil/libs_arm.go
+++ b/internal/ebutil/libs_arm.go
@@ -2,5 +2,6 @@
 
 package ebutil
 
-// Based on a 32-bit Raspbian system.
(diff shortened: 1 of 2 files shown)


johnstcn (commented):
This directory will never exist as far as I can tell on a 32-bit ARM system. I'm not sure what the wider ramifications of that would be?

johnstcn (approved):
:+1:

PR #261, opened 2024-07-02
ci: add linters, run linters, make linters happy

johnstcn:
because linters are your friend


dannykopping:
> because linters are your friend

Debatable 😆

Issue #260, opened 2024-07-02
Stop using deprecated codersdk.LogsSender function


johnstcn:
We currently use a manually-vendored version of `codersdk.LogsSender`. This is deprecated and will be removed eventually.
Instead, we should use [NewLogSender](https://pkg.go.dev/github.com/coder/coder/v2@v2.12.3/codersdk/agentsdk#NewLogSender) which uses the versioned agent API.

PR #264, opened 2024-07-04
feat: send logs using agent api v2 if available

johnstcn:
Closes https://github.com/coder/envbuilder/issues/260

* Removes manually vendored codersdk, we're back on the real deal now
* Moves logging setup to package `internal/eblog`
* Modifies logging to use new agent api v2 methods, falling back to PatchLogs
* Adds `ENVBUILDER_VERBOSE`

Tested manually against Coder version 2.8.0, 2.9.0, and 2.13.0.


go.mod
```diff
@@ -1,18 +1,20 @@
...
-
-toolchain go1.22.3
+go 1.22.4
 
 // There are a few options we need added to Kaniko!
 // See: https://github.com/GoogleContainerTools/kaniko/compare/main...coder:kaniko:main
 replace github.com/GoogleContainerTools/kaniko => github.com/coder/kaniko v0.0.0-20240624091120-7208a49f5b15
 
+// Required to import codersdk due to gvisor dependency.
+replace tailscale.com => github.com/coder/tailscale v1.1.1-0.20240702054557-aa558fbe5374
```
johnstcn, line 10: note: required to import codersdk

internal/eblog/coder.go
```diff
@@ -0,0 +1,103 @@
...
+var (
+	rpcConnectTimeout  = 10 * time.Second
+	logSendGracePeriod = 10 * time.Second
+)
+
+// Coder establishes a connection to the Coder instance located at
+// coderURL and authenticates using token. It then establishes a
+// dRPC connection to the Agent API and begins sending logs.
+// The returned function is used to block until all logs are sent.
+func Coder(ctx context.Context, coderURL *url.URL, token string) (LogFunc, func(), error) {
```
johnstcn, line 30: To remove the possibility of this breaking older users, I can check the Coder version and conditionally return a different implementation that uses the old deprecated endpoint. WDYT?

matifali: We can technically break as we are bumping a major version.

johnstcn: I ended up adding support for backward compatibility based on a similar commit in [envbox](https://github.com/coder/envbox/commit/4e0a070ae372cb1433176df815e69d7e0bcd8443).

scripts/develop.sh
```diff
@@ -3,9 +3,9 @@
...
 set -euxo pipefail
 
-./build.sh
+./build.sh || exit 1
 
 docker run --rm -it \
-    -e GIT_URL=https://github.com/denoland/deno \
-    -e INIT_SCRIPT="bash" \
+    -e ENVBUILDER_GIT_URL=https://github.com/denoland/deno \
+    -e ENVBUILDER_INIT_SCRIPT="bash" \
```
johnstcn, line 10: review: drive-by

options.go
```diff
@@ -125,7 +123,9 @@ type Options struct {
 	// execute it after successful startup.
 	PostStartScriptPath string
 	// Logger is the logger to use for all operations.
-	Logger LoggerFunc
+	Logger eblog.LogFunc
```
mtojek, line 126: nit: my brain read it as e-blog. I'm thinking whether we should stick to normal package names like `util`, `log`.

scripts/develop.sh
```diff
@@ -3,9 +3,9 @@
 cd "$(dirname "${BASH_SOURCE[0]}")"
 set -euxo pipefail
 
-./build.sh
+./build.sh || exit 1
 
 docker run --rm -it \
-    -e GIT_URL=https://github.com/denoland/deno \
-    -e INIT_SCRIPT="bash" \
+    -e ENVBUILDER_GIT_URL=https://github.com/denoland/deno \
```
mtojek, line 9: Is it a breaking change?

BTW We may need to update README file too.

internal/eblog/coder.go
```diff
@@ -0,0 +1,136 @@
...
+	retryCtx, retryCancel := context.WithTimeout(context.Background(), rpcConnectTimeout)
+	for r := retry.New(100*time.Millisecond, time.Second); r.Wait(retryCtx); {
+		// Maximize compatibility.
+		c, err = client.ConnectRPC20(ctx)
+		if err != nil {
+			continue
+		}
+		break
+	}
+	retryCancel()
```
mtojek, line 77: is it on purpose not deferred?

internal/eblog/coder_internal_test.go
```diff
@@ -0,0 +1,156 @@
...
+	"cdr.dev/slog/sloggers/slogtest"
+	"github.com/coder/coder/v2/agent/proto"
+	"github.com/coder/coder/v2/codersdk"
+	"github.com/coder/coder/v2/codersdk/agentsdk"
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_Coder(t *testing.T) {
```
mtojek, line 22: nit: TestCoder?

mtojek (approved):
Mostly nit picks

options.go
johnstcn: This was just my brain on auto-pilot, will rename :+1:

scripts/develop.sh
johnstcn: This would only be a breaking change if you were using the `develop.sh` script with a several-month-old old version of envbuilder. I don't think this counts.

johnstcn: README was updated in https://github.com/coder/envbuilder/pull/203

go.mod
dannykopping: Is the commit significant, or is it just used by virtue of being the tip of `main`?

go.mod
```diff
@@ -1,18 +1,20 @@
...
+
 require (
 	cdr.dev/slog v1.6.2-0.20240126064726-20367d4aede6
 	github.com/GoogleContainerTools/kaniko v1.9.2
 	github.com/breml/rootcerts v0.2.10
 	github.com/chainguard-dev/git-urls v1.0.2
+	github.com/coder/coder/v2 v2.10.1-0.20240704130443-c2d44d16a352
```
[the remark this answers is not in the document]
dannykopping: Interesting that it still shows `v2.10.1`.

internal/eblog/coder_internal_test.go
johnstcn: fixed

internal/log/coder.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/coder/coder/v2/codersdk"
+	"github.com/coder/coder/v2/codersdk/agentsdk"
+	"github.com/coder/retry"
+	"github.com/google/uuid"
+	"golang.org/x/mod/semver"
+)
+
+var (
+	rpcConnectTimeout  = 10 * time.Second
+	logSendGracePeriod = 10 * time.Second
```
dannykopping, line 23: Nit: I wonder if these should be configurable; at least the timeout feels like it should be.

internal/eblog/coder.go
johnstcn: fixed

internal/log/coder.go
```diff
@@ -0,0 +1,136 @@
...
+// fall back to using the PatchLogs endpoint.
+// The returned function is used to block until all logs are sent.
+func Coder(ctx context.Context, coderURL *url.URL, token string) (Func, func(), error) {
+	client := initClient(coderURL, token)
+	bi, err := client.SDK.BuildInfo(ctx)
+	if err != nil {
+		return nil, nil, fmt.Errorf("get coder build version: %w", err)
+	}
+	if semver.Compare(semver.MajorMinor(bi.Version), minAgentAPIV2) < 0 {
+		sendLogs, flushLogs := sendLogsV1(ctx, client)
```
dannykopping, line 40: Nit: I think we should log that we're using the old or the new API.

internal/log/coder.go
```diff
@@ -0,0 +1,136 @@
...
+	sendLogs, flushLogs := agentsdk.LogsSender(agentsdk.ExternalLogSourceID, client.PatchLogs, slog.Logger{})
+	return func(l Level, msg string, args ...any) {
+			log := agentsdk.Log{
+				CreatedAt: time.Now(),
+				Output:    fmt.Sprintf(msg, args...),
+				Level:     codersdk.LogLevel(l),
+			}
+			_ = sendLogs(ctx, log)
+		}, func() {
+			_ = flushLogs(ctx)
```
dannykopping, line 97: Nit: let's not drop errors without at least a debug log

internal/log/coder.go
```diff
@@ -0,0 +1,136 @@
...
+			}
+		}
+
+		// Wait for up to 10 seconds for logs to finish sending.
+		sendCtx, sendCancel := context.WithTimeout(context.Background(), logSendGracePeriod)
+		defer sendCancel()
+		// Try once more to send any pending logs
+		_ = ls.SendLoop(sendCtx, dest)
+		ls.Flush(uid)
+		_ = ls.WaitUntilEmpty(sendCtx)
```
dannykopping, line 120: Nit: probably worthwhile logging these errors too

internal/log/coder_internal_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+		// Cancel and wait for flush
+		cancel()
+		t.Logf("cancelled")
+		logsDone()
+
+		require.Len(t, ld.logs, 10)
+	})
+
+	t.Run("V2/Err", func(t *testing.T) {
```
dannykopping, line 119: I think you could express the intent of this test a bit clearer. What is causing the `initRPC` call to fail?
Checking for `init coder rpc client` is the less specific part of the error to be checking, no?

cmd/envbuilder/main_test.go
```diff
@@ -1,84 +0,0 @@
...
-
-	"cdr.dev/slog/sloggers/slogtest"
-	"github.com/coder/envbuilder/internal/notcodersdk"
-	"github.com/coder/serpent"
-	"github.com/google/uuid"
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
-)
-
-func Test_sendLogs(t *testing.T) {
```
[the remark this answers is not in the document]
dannykopping: Do we have an integration test for this?

dannykopping (approved):
LGTM, minor points

go.mod
johnstcn: same commit as from `coder/coder` repo as imported

go.mod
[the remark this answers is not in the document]
johnstcn: Yeah I noticed that too, I think it's related to the last tag on `main`:
```
git tag --merged main | sort -V | tail -1
v2.10.0
```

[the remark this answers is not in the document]
dannykopping: Except in `go.mod` it's `v2.10.1`?

cmd/envbuilder/main_test.go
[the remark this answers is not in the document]
johnstcn: No, not yet. I think this could be a good candidate for a follow-up, but it would mean we need to:
- Stand up an in-memory Coder instance
- Create an envbuilder template (we currently keep envbuilder templates in coder/coder)
- Assert the logs get sent to Coder

It would be fairly heavyweight all in all. What I'd really like is a more lightweight `fluentd`-like utility that just consumes stdin and forwards to a Coder instance.

internal/log/coder.go
johnstcn: fixed

internal/log/coder.go
johnstcn: I'll leave this as a follow up, we may not end up needing an extra knob.

internal/log/coder.go
johnstcn: fixed

internal/log/coder.go
johnstcn: fixed

internal/log/coder_internal_test.go
johnstcn: fixed

go.mod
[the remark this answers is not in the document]
johnstcn: All I care about is that the commit is present :shrug:

PR #274, opened 2024-07-22
fix(envbuilder): add /product_uuid and /product_name to IgnorePaths by default

johnstcn:
KinD adds `/product_uuid` and `/product_name` to pods. Add these to Envbuilder's default ignore list as they otherwise cause issues in KinD.

### DIFF
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -89,7 +89,7 @@ func Run(ctx context.Context, options Options) error {
 	// Once the legacy environment variables are phased out, this can be
 	// reinstated to the previous default values.
 	if len(options.IgnorePaths) == 0 {
-		options.IgnorePaths = []string{"/var/run"}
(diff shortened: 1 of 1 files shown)


envbuilder.go
```diff
@@ -89,7 +89,7 @@ func Run(ctx context.Context, options Options) error {
 	// Once the legacy environment variables are phased out, this can be
 	// reinstated to the previous default values.
 	if len(options.IgnorePaths) == 0 {
-		options.IgnorePaths = []string{"/var/run"}
+		options.IgnorePaths = []string{"/var/run", "/product_uuid", "/product_name"}
```
dannykopping, line 92: Nit: would be good to have an explanation inline as well I think?

dannykopping (approved):

mafredri (approved):

envbuilder.go
johnstcn: Fixed!

PR #278, opened 2024-07-23
chore: refactor options to separate package

johnstcn:
Extracts Options to its own package to support easier importing from separate repositories.


mafredri (approved):
I would suggest skipping the `pkg` prefix, otherwise LGTM. 👍🏻 We already have some packages, like `devcontainer` that would make sense to have on the same level IMO.

johnstcn:
> I would suggest skipping the `pkg` prefix, otherwise LGTM. 👍🏻 We already have some packages, like `devcontainer` that would make sense to have on the same level IMO.

Fair enough. I only chose `/pkg/` due to it being a commonly-used prefix, but we can strip it.

mafredri:
> Fair enough. I only chose `/pkg/` due to it being a commonly-used prefix, but we can strip it.

True, it's not totally uncommon, but it's a bit controversial. Also considered "legacy" https://x.com/bradfitz/status/1039512487538970624 by some 😄

envbuilder.go
```diff
@@ -83,68 +84,68 @@ type DockerConfig configfile.ConfigFile
 // Logger is the logf to use for all operations.
 // Filesystem is the filesystem to use for all operations.
 // Defaults to the host filesystem.
-func Run(ctx context.Context, options Options) error {
+func Run(ctx context.Context, opts options.Options) error {
```
mtojek, line 87: Can you justify why we need this change in the PR description? Not sure if having a package just for options isn't an antipattern in Go. It smells more like Java...

mtojek (commented):
The change itself LGTM, but would like to understand the reasons behind 👍

envbuilder.go
johnstcn: When we import this from a separate repo I'd rather not have to import the whole of envbuilder and instead only the things we need.

mtojek: I guess this is opinionated, but this could land in `/pkg/api`.

PR #279, opened 2024-07-23
chore: extract git operations to separate package

johnstcn:
Extracts git-related operations to its own package.


mafredri (approved):
Same no-`pkg` suggestion here, otherwise LGTM!

PR #282, opened 2024-07-26
chore: extract constants package

johnstcn:

open to naming this something else, but `constants` was what came to mind first.

### DIFF
--- a/constants/constants.go
+++ b/constants/constants.go
@@ -1,0 +1,31 @@
+package constants
+
+import (
+	"errors"
+	"path/filepath"
+)
+
+const (
+	// WorkspacesDir is the path to the directory where
+	// all workspaces are stored by default.
+	WorkspacesDir = "/workspaces"
+
+	// EmptyWorkspaceDir is the path to a workspace that has
+	// nothing going on... it's empty!
+	EmptyWorkspaceDir = WorkspacesDir + "/empty"
+
+	// MagicDir is where all envbuilder related files are stored.
+	// This is a special directory that must not be modified
+	// by the user or images.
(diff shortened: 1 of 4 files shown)


constants/constants.go
```diff
@@ -0,0 +1,31 @@
...
+
+var (
+	ErrNoFallbackImage = errors.New("no fallback image has been specified")
+
+	// MagicFile is a file that is created in the workspace
+	// when envbuilder has already been run. This is used
+	// to skip building when a container is restarting.
+	// e.g. docker stop -> docker start
+	MagicFile = filepath.Join(MagicDir, "built")
+)
```
mafredri, line 31: I do feel like this and options belong in the same package, how about merging them under `core` or something? Not sure it's better description wise but still. Ideally I'd love for these and options to be in the package named `envbuilder` since that makes the most sense. But then we either lose the separation, or need to move the Run method elsewhere.

I suppose a `build` package could be appropriate for those, but then we'd kinda revert some of the splits you've already been doing.

constants/constants.go
```diff
@@ -0,0 +1,31 @@
+package constants
```
mafredri, line 1: Two alternatives, `config` and `defaults`. Not sure if/which would be better.

mafredri (approved):
I'm not super stoked about `constants`, but not a blocker for merge. We can always revert/restructure (again) later.

constants/constants.go
johnstcn: I personally dislike both of those options :sweat_smile: but I don't have a better proposal.

PR #292, opened 2024-08-01
add support for starting envbuilder from built image

### LINKED ISSUE #291: Add support for starting envbuilder from a built image
To fully support starting envbuilder built images from cache, we need to also support starting envbuilder to setup the environment and run the init command without making changes to the filesystem.

We already have a facility for this when envbuilder detects the presence of `/.envbuilder/built` and `--skip-rebuild=true` is used. However, providing `--skip-rebuild=true` should not be necessary.

We may introduce a new indicator file (e.g. `constants.MagicImage = "/.envbuilder/image"`) to detect envbuilder built images.

mafredri:
Fixes #291

### DIFF
--- a/constants/constants.go
+++ b/constants/constants.go
@@ -28,4 +28,9 @@ var (
 	// to skip building when a container is restarting.
 	// e.g. docker stop -> docker start
 	MagicFile = filepath.Join(MagicDir, "built")
+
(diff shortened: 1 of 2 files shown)


johnstcn:
I'll integrate this into my other changes if that's OK?

PR #293, opened 2024-08-02
fix: always add COPY directives in DoCacheProbe

johnstcn:
Required in order to correctly handle RunCacheProbe. Without it, we will be unable to correctly detect the final layer of the built image.

### DIFF
--- a/options/options.go
+++ b/options/options.go
@@ -156,6 +156,11 @@ type Options struct {
 	// used to improving cache utilization when multiple users are building
 	// working on the same repository.
 	RemoteRepoBuildMode bool
+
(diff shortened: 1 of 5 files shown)


options/options.go
```diff
@@ -485,6 +497,9 @@ func (o *Options) Markdown() string {
 	_, _ = sb.WriteString("| - | - | - | - |\n")
 
 	for _, opt := range cliOptions {
+		if opt.Hidden {
```
johnstcn, line 500: review: drive-by; I don't think hidden options should be documented.

mafredri (approved):

PR #294, opened 2024-08-02
fix: only use repo mode in cache probe mode

mafredri:
Noticed we were needlessly cloning the full repo in cache probing.


johnstcn (approved):
👍

PR #296, opened 2024-08-02
feat: support starting from an already-built image

johnstcn:
Fixes https://github.com/coder/envbuilder/issues/291

Depends on https://github.com/coder/kaniko/pull/19

Builds on top of https://github.com/coder/envbuilder/pull/292
- Extracts 'magic directives' to constants.go.
- Adds 'magic image' file to signify envbuilder should skip the destructive 'build image' stage.
- Modifies logic for copying binary into built image: we now copy to build context and remove it after build finishes to avoid leaving around files owned by `root:root`. Also ensures files are created with consistent permissions.

### DIFF
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -27,6 +27,7 @@ import (
 	"github.com/coder/envbuilder/constants"
 	"github.com/coder/envbuilder/git"
 	"github.com/coder/envbuilder/options"
+	"github.com/go-git/go-billy/v5"
 
 	"github.com/GoogleContainerTools/kaniko/pkg/config"
 	"github.com/GoogleContainerTools/kaniko/pkg/creds"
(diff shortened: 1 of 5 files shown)


envbuilder.go
```diff
@@ -1382,24 +1393,32 @@ func maybeDeleteFilesystem(logger log.Func, force bool) error {
 	return util.DeleteFilesystem()
 }
 
-func copyFile(src, dst string) error {
+func copyFile(src, dst string, mode fs.FileMode) error {
 	content, err := os.ReadFile(src)
```
mafredri, line 1397: In #292 I had converted these to use billy filesystem, any reason not to do this?

envbuilder.go
```diff
@@ -354,8 +373,10 @@ ENTRYPOINT [%q]`, exePath, exePath, exePath)
 	stderrWriter, closeStderr := log.Writer(opts.Logger)
 	defer closeStderr()
 	build := func() (v1.Image, error) {
-		_, err := opts.Filesystem.Stat(constants.MagicFile)
-		if err == nil && opts.SkipRebuild {
+		defer cleanupBuildContext()
```
mafredri, line 376: I didn't consider this before, but this cleanup made me think, should we try hard(er) to make sure we don't accidentally conflict with any existing files in the buildcontext? Say, if we place these under `.envbuilder.tmp/envbuilder` and `.envbuilder.tmp/image`?

Incidentally, if the build context is R/O we're also in for a bad time but let's defer that problem 😄.

envbuilder.go
```diff
@@ -1064,23 +1066,32 @@ func RunCacheProbe(ctx context.Context, opts options.Options) (v1.Image, error)
...
 	}
 
+	// Also touch the magic file that signifies the image has been built!
+	magicImageDest := filepath.Join(buildParams.BuildContext, filepath.Base(constants.MagicImage))
+	if err := touchFile(magicImageDest, 0o755); err != nil {
+		return nil, fmt.Errorf("touch magic image file in build context: %w", err)
+	}
+	defer func() {
+		// Clean up after we're done!
+		if err := os.Remove(envbuilderBinDest); err != nil {
```
mafredri, line 1087: Should we use opts.Filesystem here?

envbuilder.go
```diff
@@ -1138,8 +1149,8 @@ ENTRYPOINT [%q]`, exePath, exePath, exePath)
 		},
 		SrcContext: buildParams.BuildContext,
 
-		// For cached image utilization, produce reproducible builds.
-		Reproducible: opts.PushImage,
+		// When performing a cache probe, always perform reproducible snapshots.
```
mafredri, line 1152: 👍🏻

envbuilder.go
johnstcn: We can use fs!

envbuilder.go
johnstcn: I like `.envbuilder.tmp`, especially for https://github.com/coder/kaniko/pull/19

envbuilder.go
```diff
@@ -1064,23 +1076,41 @@ func RunCacheProbe(ctx context.Context, opts options.Options) (v1.Image, error)
...
+	if err := touchFile(opts.Filesystem, magicImageDest, 0o755); err != nil {
+		return nil, fmt.Errorf("touch magic image file in build context: %w", err)
+	}
+	defer func() {
+		// Clean up after we're done!
+		for _, path := range []string{magicImageDest, envbuilderBinDest, magicTempDir} {
+			if err := opts.Filesystem.Remove(path); err != nil {
+				opts.Logger(log.LevelWarn, "failed to clean up magic temp dir from build context: %w", err)
+			}
+		}
```
johnstcn, line 1111: note: there is no `fs.RemoveAll` :(

envbuilder.go
```diff
@@ -354,8 +383,10 @@ ENTRYPOINT [%q]`, exePath, exePath, exePath)
 	stderrWriter, closeStderr := log.Writer(opts.Logger)
 	defer closeStderr()
 	build := func() (v1.Image, error) {
-		_, err := opts.Filesystem.Stat(constants.MagicFile)
-		if err == nil && opts.SkipRebuild {
+		defer cleanupBuildContext()
```
mafredri, line 386: There's a defer above as well, will we end up logging errors the second time?

envbuilder.go
```diff
@@ -1064,23 +1076,41 @@ func RunCacheProbe(ctx context.Context, opts options.Options) (v1.Image, error)
...
 
+	// Also touch the magic file that signifies the image has been built!
+	magicImageDest := filepath.Join(
+		magicTempDir,
+		filepath.Base(constants.MagicImage),
+	)
+	if err := touchFile(opts.Filesystem, magicImageDest, 0o755); err != nil {
+		return nil, fmt.Errorf("touch magic image file in build context: %w", err)
+	}
+	defer func() {
```
mafredri, line 1105: Does this cleanup happen too late? If we run into an error we will dirty the context?

mafredri (approved):

envbuilder.go
johnstcn: Good point, I wrapped it in a `sync.Once`.

envbuilder.go
johnstcn: Moved it right up.

PR #298, opened 2024-08-05
feat: embed version info into binary

johnstcn:
Adds version info to envbuilder binary. Most of the logic is fork-lifted from `coder/coder`.

### DIFF
--- a/scripts/version.sh
+++ b/scripts/version.sh
@@ -27,12 +27,12 @@ if [[ -z ${tag_list} ]]; then
 	log
 	log "INFO(version.sh): It appears you've checked out a fork or shallow clone of Envbuilder."
 	log "INFO(version.sh): By default GitHub does not include tags when forking."
-	log "INFO(version.sh): We will use the default version 2.0.0 for this build."
+	log "INFO(version.sh): We will use the default version 0.0.1 for this build."
 	log "INFO(version.sh): To pull tags from upstream, use the following commands:"
 	log "INFO(version.sh):   - git remote add upstream https://github.com/coder/envbuilder.git"
 	log "INFO(version.sh):   - git fetch upstream"
 	log
-	last_tag="v2.0.0"
+	last_tag="v0.0.1"
 else
 	current_commit=$(git rev-parse HEAD)
 	# Try to find the last tag that contains the current commit
@@ -48,7 +48,7 @@ version="${last_tag}"
 
 # If the HEAD has extra commits since the last tag then we are in a dev version.
 #
-# Dev versions are denoted by the "-devel+" suffix with a trailing commit short
+# Dev versions are denoted by the "-dev+" suffix with a trailing commit short
 # SHA.
 if [[ "${ENVBUILDER_RELEASE:-}" == *t* ]]; then
 	# $last_tag will equal `git describe --always` if we currently have the tag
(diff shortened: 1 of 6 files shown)


scripts/version.sh
```diff
@@ -1,10 +1,77 @@
...
+fi
+
+# To make contributing easier, if there are no tags, we'll use a default
+# version.
+tag_list=$(git tag)
+if [[ -z ${tag_list} ]]; then
+	log
+	log "INFO(version.sh): It appears you've checked out a fork or shallow clone of Envbuilder."
+	log "INFO(version.sh): By default GitHub does not include tags when forking."
+	log "INFO(version.sh): We will use the default version 2.0.0 for this build."
```
dannykopping, line 30: Is v2.0.0 legit?

dannykopping (approved):

scripts/version.sh
```diff
@@ -1,10 +1,77 @@
...
+	log "INFO(version.sh): To pull tags from upstream, use the following commands:"
+	log "INFO(version.sh):   - git remote add upstream https://github.com/coder/envbuilder.git"
+	log "INFO(version.sh):   - git fetch upstream"
+	log
+	last_tag="v2.0.0"
```
mafredri, line 35: Fixme?

scripts/version.sh
mafredri, line 30: Do we use 0.0.1 for envbuilder, perhaps?

scripts/version.sh
```diff
@@ -1,10 +1,77 @@
...
+	if [[ -z "${last_tag}" ]]; then
+		last_tag=$(git tag --sort=version:refname | tail -n 1)
+	fi
+fi
+
+version="${last_tag}"
+
+# If the HEAD has extra commits since the last tag then we are in a dev version.
+#
+# Dev versions are denoted by the "-devel+" suffix with a trailing commit short
```
mafredri, line 51: We shouldn't inherit `-devel+` from coder/coder, it's not totally semver compatible. Preferable would be `+devel-commit` or something like that.

scripts/version.sh
```diff
@@ -1,10 +1,77 @@
...
+else
+	version+="-dev-$(git rev-parse --short HEAD)"
+fi
 
-last_tag="$(git describe --tags --abbrev=0)"
-version="$last_tag"
+# If the git repo has uncommitted changes, mark the version string as 'dirty'.
+dirty_files=$(git ls-files --other --modified --exclude-standard)
+if [[ -n "${dirty_files}" ]]; then
+	version+="-dirty"
```
mafredri, line 73:
```suggestion
	version+="+dirty"
```

`-` is reserved for release type, e.g. `alpha`. `+` is for build information.

https://semver.org

mafredri (approved):
We should probably strip out more cruft that we don't want to inherit, but no major blockers for now so approving.

scripts/version.sh
johnstcn: nope, copypasta

scripts/version.sh
johnstcn: v1.0.0, wdyt?

scripts/version.sh
johnstcn: It seems like a valid pre-release identifier to me?

PR #303, opened 2024-08-07
ci: fix version script and update release.yaml

johnstcn:
Fixes some issues in the version script related to dev versions.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -19,7 +19,9 @@ jobs:
     name: Build and publish
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
+        with:
+          fetch-tags: true
 
       - name: Echo Go Cache Paths
         id: go-cache-paths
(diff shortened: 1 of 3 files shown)


scripts/version.sh
```diff
@@ -54,25 +54,17 @@ if [[ "${ENVBUILDER_RELEASE:-}" == *t* ]]; then
 	# $last_tag will equal `git describe --always` if we currently have the tag
 	# checked out.
 	if [[ "${last_tag}" != "$(git describe --always)" ]]; then
-		# make won't exit on $(shell cmd) failures, so we have to kill it :(
```
johnstcn, line 57: review: I don't think we need this given how we're building

scripts/lib.sh
```diff
@@ -39,3 +39,9 @@ cdroot() {
 log() {
 	echo "$*" 1>&2
 }
+
+# error prints an error message and returns an error exit code.
```
johnstcn, line 43: review: missed copy-pasta

.github/workflows/release.yaml
```diff
@@ -44,11 +46,18 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Build and Push
+      - name: Get version
+        id: get-version
+        env:
+          ENVBUILDER_RELEASE: "t"
```
johnstcn, line 52: review: the updated script will emit `-dev` versions unless this is explicitly set. In this case it will validate that the version is an annotated tag.

mafredri (approved):

dannykopping (approved):

PR #304, opened 2024-08-08
fix(scripts): correct output of version.sh

johnstcn:
We had been tagging all dev versions as v0 even though there is an annotated tag on `main`.
Also updates the release workflow to enforce the annotated tag checking.


johnstcn:
duplicate of #303

Issue #314, opened 2024-08-13
bug: devcontainer.Compile references global variable


johnstcn:
If you try to compile a Dockerfile from a devcontainer.json as non-root, you will get this error:

> Error: compile devcontainer.json: open dockerfile: mkdir /.envbuilder: permission denied

(Alternatively, you may also get a "read-only filesystem" error.)

We should instead write the Dockerfile into the build context and not just use MagicDir.

PR #315, opened 2024-08-14
fix(envbuilder): RunCacheProbe: remove references to constants.MagicDir

johnstcn:
Relates to https://github.com/coder/envbuilder/issues/314

`RunCacheProbe` is likely to execute without root permissions (e.g. via Terraform provider).
Writing to `constants.MagicDir` will fail in this case due to lack of permission.
Instead, use `buildTimeWorkspaceFolder` to write a `Dockerfile` when compiling from a `devcontainer.json`.

### DIFF
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -948,7 +948,7 @@ func RunCacheProbe(ctx context.Context, opts options.Options) (v1.Image, error) 
 	}
 
 	defaultBuildParams := func() (*devcontainer.Compiled, error) {
-		dockerfile := filepath.Join(constants.MagicDir, "Dockerfile")
+		dockerfile := filepath.Join(buildTimeWorkspaceFolder, "Dockerfile")
 		file, err := opts.Filesystem.OpenFile(dockerfile, os.O_CREATE|os.O_WRONLY, 0o644)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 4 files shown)


envbuilder.go
```diff
@@ -1010,7 +1010,7 @@ func RunCacheProbe(ctx context.Context, opts options.Options) (v1.Image, error)
 					opts.Logger(log.LevelInfo, "No Dockerfile or image specified; falling back to the default image...")
 					fallbackDockerfile = defaultParams.DockerfilePath
 				}
-				buildParams, err = devContainer.Compile(opts.Filesystem, devcontainerDir, constants.MagicDir, fallbackDockerfile, opts.WorkspaceFolder, false, os.LookupEnv)
+				buildParams, err = devContainer.Compile(opts.Filesystem, devcontainerDir, buildTimeWorkspaceFolder, fallbackDockerfile, opts.WorkspaceFolder, false, os.LookupEnv)
```
mafredri, line 1013: Will this potentially change any of the Docker instructions/"directives"? That could lead to a cache miss if that's the case. Otherwise LGTM.

mafredri (approved):

envbuilder.go
johnstcn: I see a comment in `devcontainer.go` that mentions `featureDir` must be reproducible based on the feature name, but I think that's relative to the build context.

In any case, I think it makes sense to beef up the testing around `RunCacheProbe` outside of a container as a non-root user.

johnstcn: I looked into adding some more non-rootful cache probe tests and I then remembered that you actually need the envbuilder binary for this. The infrastructure for all that is already there in the terraform provider tests, so that's probably a better place to do it.

PR #336, opened 2024-09-06
fix: set DOCKER_CONFIG dynamically

johnstcn:
Relates to https://github.com/coder/terraform-provider-envbuilder/issues/43

We were not setting `DOCKER_CONFIG` correctly, and were instead relying on having it set for us in the image. This change dynamically sets `DOCKER_CONFIG` at runtime instead, and removes the definition in the Envbuilder image.

### DIFF
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -1475,7 +1475,7 @@ func initDockerConfigJSON(logf log.Func, magicDir magicdir.MagicDir, dockerConfi
 	if dockerConfigBase64 == "" {
 		return noop, nil
 	}
-	cfgPath := filepath.Join(magicDir.Path(), "config.json")
+	cfgPath := magicDir.Join("config.json")
 	decoded, err := base64.StdEncoding.DecodeString(dockerConfigBase64)
 	if err != nil {
 		return noop, fmt.Errorf("decode docker config: %w", err)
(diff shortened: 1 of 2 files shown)


scripts/Dockerfile
```diff
@@ -4,7 +4,5 @@ ARG TARGETARCH
 COPY envbuilder-${TARGETARCH} /.envbuilder/bin/envbuilder
 
 ENV KANIKO_DIR /.envbuilder
-# Kaniko looks for the Docker config at $DOCKER_CONFIG/config.json
-ENV DOCKER_CONFIG /.envbuilder
```
johnstcn, line 8: This may seem controversial, but I think it definitely makes sense to set it in code if we allow dynamically changing `MagicDir`.

envbuilder.go
```diff
@@ -1490,9 +1490,15 @@ func initDockerConfigJSON(logf log.Func, magicDir constants.MagicDir, dockerConf
 		return noop, fmt.Errorf("write docker config: %w", err)
 	}
 	logf(log.LevelInfo, "Wrote Docker config JSON to %s", cfgPath)
+	oldDockerConfig := os.Getenv("DOCKER_CONFIG")
+	os.Setenv("DOCKER_CONFIG", magicDir.String())
+	logf(log.LevelInfo, "Set DOCKER_CONFIG to %s", magicDir.String())
```
dannykopping, line 1495: Nit: for the absence of doubt, it might be best to read the newly-set env var instead of what it _should_ (and almost certainly _will_) be.

dannykopping (approved):

mtojek (approved):
stamp 👍

mafredri (approved):

PR #337, opened 2024-09-06
fix: allow setting MagicDir in Options

johnstcn:
Relates to https://github.com/coder/terraform-provider-envbuilder/issues/43

This PR unexports `constants.MagicDir` and replaces it with a different implementation that supports dynamically changing `MagicDir` at runtime.
There is also some more "strategic" hard-coding of certain values to ensure they do not change when `MagicDir` changes.

See also: https://github.com/coder/envbuilder/pull/336 

A subsequent PR will update the Terraform provider.


constants/constants.go
```diff
@@ -51,14 +27,46 @@ var (
...
+COPY --chmod=0644 .envbuilder.tmp/image /.envbuilder/image
 USER root
 WORKDIR /
-ENTRYPOINT [%[2]q]
-`,
-		".envbuilder.tmp/envbuilder", MagicBinaryLocation,
-		".envbuilder.tmp/image", MagicImage,
-	)
+ENTRYPOINT ["/.envbuilder/bin/envbuilder"]
+`
```
johnstcn, line 36: review: Hard-coding the values here, as they should never change.

constants/constants.go
```diff
@@ -51,14 +27,46 @@ var (
...
+	return filepath.Join(m.String(), "built")
+}
+
+// MagicImage is a file that is created in the image when
+// envbuilder has already been run. This is used to skip
+// the destructive initial build step when 'resuming' envbuilder
+// from a previously built image.
+func (m MagicDir) Image() string {
+	return filepath.Join(m.String(), "image")
+}
```
johnstcn, line 72: I'm not particularly married to this implementation, but it just seemed like a nice way to do it.

envbuilder.go
```diff
@@ -1417,21 +1407,24 @@ func findDevcontainerJSON(workspaceFolder string, options options.Options) (stri
...
 			logger(log.LevelError, "To bypass this check, set FORCE_SAFE=true.")
 			return errors.New("safety check failed")
 		}
+		bailoutSecs := 10
+		logger(log.LevelWarn, "WARNING! BYPASSING SAFETY CHECK! THIS WILL DELETE YOUR ROOT FILESYSTEM!")
+		logger(log.LevelWarn, "You have %d seconds to bail out!", bailoutSecs)
+		for i := bailoutSecs; i > 0; i-- {
+			logger(log.LevelWarn, "%d...", i)
+			<-time.After(time.Second)
+		}
```
johnstcn, line 1427: review: inverted the logic here for readability.

envbuilder.go
```diff
@@ -1489,10 +1482,14 @@ func initDockerConfigJSON(dockerConfigBase64 string) (func() error, error) {
 	if err != nil {
 		return noop, fmt.Errorf("parse docker config: %w", err)
 	}
+	for k := range configFile.AuthConfigs {
+		logf(log.LevelInfo, "Docker config contains auth for registry %q", k)
+	}
```
johnstcn, line 1487: review: I found this useful when troubleshooting!

integration/integration_test.go
```diff
@@ -1596,6 +1599,111 @@ COPY --from=prebuild /the-future/hello.txt /the-future/hello.txt
...
+	t.Run("MultistageRunCopyMinimal", func(t *testing.T) {
+		t.Parallel()
+
+		srv := gittest.CreateGitServer(t, gittest.Options{
+			Files: map[string]string{
+				"Dockerfile": fmt.Sprintf(`
+FROM %[1]s AS a
+RUN date > /date.txt
+FROM %[1]s
+COPY --from=a /date.txt /date.txt
```
johnstcn, line 1611: I didn't actually mean to include this test; I can remove it if folks feel strongly. It doesn't really expand on anything from the above "hello from the future" test.

options/options.go
```diff
@@ -166,6 +167,11 @@ type Options struct {
 	// attempting to probe the build cache. This is only relevant when
 	// GetCachedImage is true.
 	BinaryPath string
+
+	// MagicDir is the path to the directory where all envbuilder files should be
+	// stored. By default, this is set to `/.envbuilder`. This is intentionally
+	// excluded from the CLI options.
+	MagicDir constants.MagicDir
```
johnstcn, line 174: review: you can now set MagicDir in Options. You can't set it through `ENVBUILDER_MAGIC_DIR` though; this is intentional.

constants/constants.go
dannykopping: Nit: I think we should still use the above `defaultMagicDir` const.
They are ostensibly linked conceptually, and if we changed one without the other it could cause problems right?

constants/constants.go
dannykopping: Nit: perhaps rename `String()` to `Path()` and have `Image()` and a new method `String()` both call `Path()` to be more explicit.

envbuilder.go
```diff
@@ -332,31 +332,26 @@ func Run(ctx context.Context, opts options.Options) error {
...
-			magicTempDir,
-			filepath.Base(constants.MagicBinaryLocation),
-		)
+		envbuilderBinDest := filepath.Join(magicTempDir.String(), "envbuilder")
 		// Also touch the magic file that signifies the image has been built!
-		magicImageDest := filepath.Join(
-			magicTempDir,
-			filepath.Base(constants.MagicImage),
-		)
+		// magicImageDest := filepath.Join(magicTempDir, "image")
```
dannykopping, line 348: Nit: remove?

constants/constants.go
```diff
@@ -51,14 +27,46 @@ var (
...
+
+// MagicDir implements fmt.Stringer.
+var _ fmt.Stringer = MagicDir("")
+
+// MagicFile is a file that is created in the workspace
+// when envbuilder has already been run. This is used
+// to skip building when a container is restarting.
+// e.g. docker stop -> docker start
+func (m MagicDir) Built() string {
```
dannykopping, line 62: Bit: I feel like this should return `bool`, and have another method `MarkBuilt()` or something which tries to create the path. I feel like this is currently bleeding implementation details out.

envbuilder.go
```diff
@@ -1501,7 +1498,7 @@ func initDockerConfigJSON(dockerConfigBase64 string) (func() error, error) {
 				if !errors.Is(err, fs.ErrNotExist) {
 					cleanupErr = fmt.Errorf("remove docker config: %w", cleanupErr)
 				}
-				_, _ = fmt.Fprintf(os.Stderr, "failed to remove the Docker config secret file: %s\n", cleanupErr)
+				logf(log.LevelError, "failed to remove the Docker config secret file: %s", cleanupErr)
```
dannykopping, line 1501: Consistency nit: looks like all other log entries in this PR start with a capital letter.
```suggestion
				logf(log.LevelError, "Failed to remove the Docker config secret file: %s", cleanupErr)
```

integration/integration_test.go
```diff
@@ -365,7 +365,8 @@ func TestBuildFromDockerfile(t *testing.T) {
 	require.Equal(t, "hello", strings.TrimSpace(output))
 
 	// Verify that the Docker configuration secret file is removed
-	output = execContainer(t, ctr, "stat "+filepath.Join(constants.MagicDir, "config.json"))
+	configJSONContainerPath := filepath.Join(constants.MagicDir("").String(), "config.json")
```
dannykopping, line 368: Idea: instead of `constants.MagicDir("")` why not add a func like `constants.DefaultMagicDir()`?

dannykopping (approved):
LGTM, only nits.

mtojek (approved):
👍

constants/constants.go
```diff
@@ -51,14 +27,46 @@ var (
...
+type MagicDir string
+
+// String returns the string representation of the MagicDir.
+func (m MagicDir) String() string {
+	if m == "" {
+		// Instead of the zero value, use defaultMagicDir.
+		return defaultMagicDir
+	}
+	return filepath.Join("/", string(m))
+}
```
mafredri, line 53: I'm thinking this abstraction could be a source of bugs. One might not always think to call `String` and instead do `string(MagicDir)` which would result in an empty string or potentially lack of `/`.

I'd rather see this as a validation step in options so that the input is verified there, and an error can be returned if it's not absolute. The options struct could have the `MagicImage` and `MagicBuilt` functions if that's preferable.

mafredri (commented):
I feel the `constants` package is becoming pretty useless with the dynamic type. What do you think about moving more of the logic to options instead?

constants/constants.go
johnstcn: I think a simpler option here it to simply change the type to a struct with a single unexported field. I'm less worried about leaking the implementation details; we can simply move `MagicDir` to an internal package.

constants/constants.go
johnstcn: While this is true, I think this can be mostly addressed by simply moving it to an internal package.

constants/constants.go
johnstcn: Updated!

constants/constants.go
johnstcn: Done, except I got rid of the `String()` entirely to avoid accidental stringification. I did add a `Join()` convenience method though.

johnstcn:
> I feel the `constants` package is becoming pretty useless with the dynamic type. What do you think about moving more of the logic to options instead?

I moved some of the logic to options, and the rest to `internal/magicdir`.

Issue #281, opened 2024-07-24
envbuilder caches repo 


reiser:
Is it possible to disable the /workspace cache of a coder build with envbuilder.
Even after restart work spaces it caches to old commit

```
#1: 📦 Cloning https://bitbucket.org/lorem/repo.git to /workspaces/ipsum..
#1: 👤 Using no authentication!
#1: 📦 The repository already exists! [672.333µs]
```

johnstcn:
What should envbuilder do in this case if the repository contains uncommitted changes?
Should it still fetch changes from the remote?

megla-tlanghorst:
Couldn't it check for local changes? I had this when the devcontainer build failed and I fixed it in the repo, but I had to destroy the workspace because I couldn't update it easily obviously.

mafredri:
@megla-tlanghorst would the [`--remote-repo-build-mode`](https://github.com/coder/envbuilder?tab=readme-ov-file#environment-variables) option we recently introduced cover your use-case?

Automatically updating the users repo is an interesting proposition too, though. There are some caveats though:

- What do we consider dirty? Uncommitted changes? Untracked files?
- What do we do if there's an error? I.e. untracked file -> pull -> conflict
- What if the user was working on a specific feature but since we pulled the changes they're no longer sure where they were (if it's an opt-in option, the onus is on the user, but it's still a potentially bad experience)

megla-tlanghorst:
> @megla-tlanghorst would the [`--remote-repo-build-mode`](https://github.com/coder/envbuilder?tab=readme-ov-file#environment-variables) option we recently introduced cover your use-case?

Maybe, that's what I'm currently testing.

johnstcn:
Alternative possibility is add -> stash -> pull -> unstash. This does run the risk of users being left to clean up a nasty merge conflict though.

johnstcn:
Parking this for now due to some open questions:
- Do we `fetch` and `checkout` versus `pull --ff-only`? The former seems safer to me.
- What do we do in case of local changes? It seems reasonable to just skip fetching updates completely if the worktree is dirty, but is this the behaviour that users expect?
- What should we do if a user attempts to clone a different Git repo into an existing one? Right now, we just skip cloning altogether if we detect an existing repo. Should we update the existing repo, or bail entirely?
- What if the pre-existing repo has a different remote configured? Which remote do we use? The same as `ENVBUILDER_GIT_URL`?

mafredri:
[...]

PR #345, opened 2024-09-12
feat(git): log parsed gitURL and warn if local

johnstcn:
Part of https://github.com/coder/envbuilder/issues/281

If we go about fetching changes from a remote repo, we're going to want to log some more information about what's going on.
I also took this opportunity to log some more information about the git URL we parse.
Specifically, if a user passes a git URL in the form of `OrgName/RepoName` our library for parsing git URLs will interpret that as a local URL (`file://`). 
`go-git` will fall back to the `git` executable in this case, which is likely to fail if `git` is not present. I'm choosing to log a warn but not block in case someone decides to make their own envbuilder image with `git` installed.

### DIFF
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -112,22 +112,25 @@ func Run(ctx context.Context, opts options.Options) error {
 	var fallbackErr error
 	var cloned bool
 	if opts.GitURL != "" {
-		cloneOpts, err := git.CloneOptionsFromOptions(opts)
+		endStage := startStage("📦 Cloning %s to %s...",
+			newColor(color.FgCyan).Sprintf(opts.GitURL),
+			newColor(color.FgCyan).Sprintf(opts.WorkspaceFolder),
+		)
+		stageNum := stageNumber
+		logStage := func(format string, args ...any) {
+			opts.Logger(log.LevelInfo, "#%d: %s", stageNum, fmt.Sprintf(format, args...))
+		}
+
+		cloneOpts, err := git.CloneOptionsFromOptions(logStage, opts)
 		if err != nil {
 			return fmt.Errorf("git clone options: %w", err)
 		}
 
-		endStage := startStage("📦 Cloning %s to %s...",
-			newColor(color.FgCyan).Sprintf(opts.GitURL),
-			newColor(color.FgCyan).Sprintf(cloneOpts.Path),
-		)
-
-		stageNum := stageNumber
-		w := git.ProgressWriter(func(line string) { opts.Logger(log.LevelInfo, "#%d: %s", stageNum, line) })
+		w := git.ProgressWriter(func(line string) { logStage(line) })
 		defer w.Close()
 		cloneOpts.Progress = w
 
-		cloned, fallbackErr = git.CloneRepo(ctx, cloneOpts)
+		cloned, fallbackErr = git.CloneRepo(ctx, logStage, cloneOpts)
 		if fallbackErr == nil {
 			if cloned {
 				endStage("📦 Cloned repository!")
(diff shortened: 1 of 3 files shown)


git/git.go
```diff
@@ -40,6 +40,7 @@ type CloneRepoOptions struct {
 	Depth        int
 	CABundle     []byte
 	ProxyOptions transport.ProxyOptions
+	Logf         log.Func
```
mafredri, line 43: Let's change this to `func(line string)` (or printf signature if you prefer).

I don't want us logging `#1` in the clone (or package for that matter) as it should not have any context of where/when it runs.

We should probably fix other instances of similar behavior as well.

johnstcn: @mafredri if I understand correctly, your idea is to use the return value of `startStage()` instead? I think that's reasonable.

envbuilder.go
```diff
@@ -112,22 +112,25 @@ func Run(ctx context.Context, opts options.Options) error {
...
+		logStage := func(format string, args ...any) {
+			opts.Logger(log.LevelInfo, "#%d: %s", stageNum, fmt.Sprintf(format, args...))
+		}
+
+		cloneOpts, err := git.CloneOptionsFromOptions(logStage, opts)
+		if err != nil {
+			return fmt.Errorf("git clone options: %w", err)
+		}
+
+		w := git.ProgressWriter(func(line string) { logStage(line) })
```
mafredri, line 129: Maybe we just change the progress writer signature to match, saves us from the function wrapping, wdyt?

git/git.go
```diff
@@ -47,11 +46,12 @@ type CloneRepoOptions struct {
 // be cloned again.
 //
 // The bool returned states whether the repository was cloned or not.
-func CloneRepo(ctx context.Context, opts CloneRepoOptions) (bool, error) {
+func CloneRepo(ctx context.Context, logf func(string, ...any), opts CloneRepoOptions) (bool, error) {
```
mafredri, line 49: I think this could be optional as part of options, but not a blocker.

git/git.go
```diff
@@ -182,14 +183,14 @@ func ReadPrivateKey(path string) (gossh.Signer, error) {
...
-func LogHostKeyCallback(logger log.Func) gossh.HostKeyCallback {
+func LogHostKeyCallback(logger func(string, ...any)) gossh.HostKeyCallback {
 	return func(hostname string, remote net.Addr, key gossh.PublicKey) error {
 		var sb strings.Builder
 		_ = knownhosts.WriteKnownHost(&sb, hostname, remote, key)
 		// skeema/knownhosts uses a fake public key to determine the host key
 		// algorithms. Ignore this one.
 		if s := sb.String(); !strings.Contains(s, "fake-public-key ZmFrZSBwdWJsaWMga2V5") {
-			logger(log.LevelInfo, "#1: 🔑 Got host key: %s", strings.TrimSpace(s))
+			logger("🔑 Got host key: %s", strings.TrimSpace(s))
```
mafredri, line 193: ❤️

mafredri (approved):
Nice, thanks for fixing a bunch of old inconsistencies!

envbuilder.go
johnstcn: Yep just realised that! Done. Also forgot that `%` in git output will end up `%(MISSING)` 😅

git/git.go
johnstcn: I don't think it is optional! If it's not provided, we'll try to log and panic.

mafredri: Oh, it would be via `!= nil` check or assigning a noop function.

PR #350, opened 2024-09-19
chore: update dind examples to use onCreateCommand

johnstcn:
The existing Docker examples override `ENVBUILDER_INIT_SCRIPT`, which is annoying if you're trying to use them with Coder. Update the examples in `docker` to use `onCreateCommand` instead.

To test them out locally:

```
make build
docker run -it --rm --privileged -v $PWD/examples/docker/02_dind:/workspaces/empty -e ENVBUILDER_INIT_SCRIPT=bash envbuilder:latest
```


examples/docker/02_dind/on-create.sh
```diff
@@ -0,0 +1,22 @@
...
+  fi
+  if [[ ! -e /var/run/docker.sock ]]; then
+    sleep 1
+  else
+    break
+  fi
+done
+# Change the owner of the Docker socket so that the coder user can use it.
+# Using `newgrp docker` is kind of annoying.
+sudo chown coder:coder /var/run/docker.sock
```
matifali, line 22: Can this cause issues with other services in the workspace which rely on docker group membership to access the docker socket?

johnstcn: Good point, let's keep the group ownership as `docker`.

examples/docker/02_dind/on-create.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# Start Docker in the background
+sudo -u root /bin/sh -c 'nohup dockerd 2>&1 > /var/log/docker.log &'
```
mafredri, line 6:
```suggestion
sudo -u root /bin/sh -c 'nohup dockerd > /var/log/docker.log &'
```
Nohup already redirects stderr and the order here was probably wrong, since `2>&1` before `1> file`, stderr would still go to the old stdout.

docs/docker.md
```diff
@@ -75,7 +75,7 @@ docker run -it --rm \
 The above can also be accomplished using the [`docker-in-docker` Devcontainer
 feature](https://github.com/devcontainers/features/tree/main/src/docker-in-docker).
 
-> Note: we still need the custom entrypoint to start the docker startup script.
+> Note: we still need `onCreateCommand` to start the docker startup script.
```
mafredri, line 78:
```suggestion
> Note: we still need the `onCreateCommand` to start Docker.
```

examples/docker/02_dind/on-create.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# Wait for Docker to start
```
mafredri, line 8:
```suggestion
# Wait up to 10 seconds for Docker to start.
```

examples/docker/02_dind/on-create.sh
```diff
@@ -0,0 +1,22 @@
...
+    exit 1
+  fi
+  if [[ ! -e /var/run/docker.sock ]]; then
+    sleep 1
+  else
+    break
+  fi
+done
+# Change the owner of the Docker socket so that the coder user can use it.
+# Using `newgrp docker` is kind of annoying.
```
mafredri, line 21:
```suggestion
```
I think we can skip this, it's already mentioned earlier.

examples/docker/03_dind_feature/Dockerfile
```diff
@@ -1,3 +1,18 @@
...
+      apt-transport-https && \
+    echo "ubuntu ALL=(ALL) NOPASSWD:ALL" > /etc/sudoers.d/ubuntu
+# Add our onCreateCommand script.
+ADD on-create.sh /on-create.sh
+# Switch to the non-root user.
+USER ubuntu
+# The devcontainer feature provides /usr/local/share/docker-init.sh
+# which will handle most of the steps of setting up Docker.
+# We can't put this in the entrypoint as it gets overridden, so
+# we call it in the on-create script.
```
mafredri, line 17: Do we not support setting entrypoint in the Dockerfile at all? Maybe we should in the future.

mafredri (approved):
Some minor nits but otherwise LGTM.

examples/docker/03_dind_feature/Dockerfile
johnstcn: My observation is that whatever entrypoint you set gets overridden by `ENVBUILDER_INIT_COMMAND` / `ENVBUILDER_INIT_SCRIPT`.

examples/docker/02_dind/on-create.sh
johnstcn: I have stories to tell you about how this can go wrong spectacularly!

examples/docker/03_dind_feature/Dockerfile
matifali: For a Coder workspace we need to have `coder_agent.*.init_script` as the entry point so another option to handle the `ENTRYPOINT` is to run it as part of `coder_agent` `startup_script`. This will work for at least the case when envbuilder is used with Coder.

johnstcn: Filed https://github.com/coder/envbuilder/issues/351 to follow up on this.

PR #355, opened 2024-09-25
fix: allow caching run layers with no filesystem changes

### LINKED ISSUE #326: Allow cache probing to work when instructions (e.g. `RUN`) don't change any files
Related: #230

In https://github.com/coder/kaniko/pull/25 we added support for multi-stage cache probing. While testing the implementation we noticed that sometimes an instruction (e.g. `RUN systemctl enable ssh`) doesn't change any files. This means the snapshotter won't create a tar archive and the cache layer won't be pushed to the remote registry.

This is problematic when running the cache probe as these instructions will not have a matching cache.

We also can't assume that a missing cache is OK because the hash of the instruction does not affect the next instruction (bug?). Essentially the instruction could be modified (`RUN systemctl enable ssh` => `RUN systemctl disable ssh`) and we'd still have a cache hit if we assumed a missing cache was OK.

Potential solutions to fix the problem:

1. Push empty cache layers for instructions with no-change
2. Allow `md5(RUN systemctl enable ssh)` to be part of the next instructions cache key

mafredri:
Fixing this issue turned out to be quite easy as Kaniko had an option for it already. We now enable caching of empty layers by enabling the `ForceBuildMetadata` option whenever push image is set.

Fixes #326

### DIFF
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -456,17 +456,18 @@ func Run(ctx context.Context, opts options.Options) error {
 			}
 			kOpts := &config.KanikoOptions{
 				// Boilerplate!
-				CustomPlatform:    platforms.Format(platforms.Normalize(platforms.DefaultSpec())),
-				SnapshotMode:      "redo",
-				RunV2:             true,
-				RunStdout:         stdoutWriter,
-				RunStderr:         stderrWriter,
-				Destinations:      destinations,
-				NoPush:            !opts.PushImage || len(destinations) == 0,
-				CacheRunLayers:    true,
-				CacheCopyLayers:   true,
-				CompressedCaching: true,
-				Compression:       config.ZStd,
+				CustomPlatform:     platforms.Format(platforms.Normalize(platforms.DefaultSpec())),
+				SnapshotMode:       "redo",
+				RunV2:              true,
+				RunStdout:          stdoutWriter,
+				RunStderr:          stderrWriter,
+				Destinations:       destinations,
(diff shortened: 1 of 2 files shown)


envbuilder.go
```diff
@@ -456,17 +456,18 @@ func Run(ctx context.Context, opts options.Options) error {
...
+				CustomPlatform:     platforms.Format(platforms.Normalize(platforms.DefaultSpec())),
+				SnapshotMode:       "redo",
+				RunV2:              true,
+				RunStdout:          stdoutWriter,
+				RunStderr:          stderrWriter,
+				Destinations:       destinations,
+				NoPush:             !opts.PushImage || len(destinations) == 0,
+				CacheRunLayers:     true,
+				CacheCopyLayers:    true,
+				ForceBuildMetadata: opts.PushImage, // Force layers with no changes to be cached, required for cache probing.
```
johnstcn, line 468: Is there any downside to always enabling this option regardless of `PushImage`?

mafredri: I'm not sure. It'll be an additional cache layer to download during builds vs executing the command and ending up with the same result. Mainly added it to be cautious.

Imagine the following scenario:

```dockerfile
RUN if [[ $(date +%s) -gt 1734998400 ]]; then touch "/'member-christmas?"; fi
```

With force build metadata, we would still use the cache here and not produce a new image.

johnstcn: Fair enough, let's leave as-is for now 👍

johnstcn (approved):

PR #359, opened 2024-09-26
fix: set env and run scripts when starting cached image

### LINKED ISSUE #358: Startup environment issues when starting a cached image
When starting a **cached image**, there is currently a few issues we need to iron out:

- [x] We don't sanitize the environment (envbuilder options are present)
- [x] Remote repo build option triggers a needless clone at startup
- [x] We skip execution of some lifecycle scripts

This list may not be exhaustive and there may be additional edge cases which need to be handled.

Issues that are present even in the absence of cached image, but can be relevant for a smooth user experience:

- [ ] If repo is present, we don't fix permissions even if repo is owned by root:root and we're running as a user (this happened to one of my development workspaces)

mafredri:
Fixes #358


envbuilder.go
```diff
@@ -58,32 +58,65 @@ var ErrNoFallbackImage = errors.New("no fallback image has been specified")
...
 // Filesystem is the filesystem to use for all operations.
 // Defaults to the host filesystem.
 func Run(ctx context.Context, opts options.Options) error {
-	defer options.UnsetEnv()
-	if opts.GetCachedImage {
-		return fmt.Errorf("developer error: use RunCacheProbe instead")
+	var args execArgsInfo
+	// Run in a separate function to ensure all defers run before we
+	// setuid or exec.
+	err := run(ctx, opts, &args)
```
mafredri, line 91: Review: Even though we've fixed defers here, there's still `cmd/envbuilder` where e.g. `closeLogs` won't fire since we execve(2) after this.

envbuilder.go
```diff
@@ -108,6 +159,32 @@ func Run(ctx context.Context, opts options.Options) error {
 		}
 	}() // best effort
 
+	runtimeData := runtimeDataStore{
+		ContainerEnv: make(map[string]string),
+		RemoteEnv:    make(map[string]string),
+	}
```
mafredri, line 165: Review: Decided to move all variables with larger scope to runtimeData, even though we only serialize container user into e.g. magic image. This is to make it more clear that some variables affect a greater scope, and allows us to share logic between both build and start from cached image.

envbuilder.go
```diff
@@ -175,112 +252,108 @@ func Run(ctx context.Context, opts options.Options) error {
...
+						opts.Logger(log.LevelInfo, "No Dockerfile or image specified; falling back to the default image...")
+						fallbackDockerfile = defaultParams.DockerfilePath
+					}
+					buildParams, err = devContainer.Compile(opts.Filesystem, devcontainerDir, magicDir.Path(), fallbackDockerfile, opts.WorkspaceFolder, false, os.LookupEnv)
+					if err != nil {
+						return fmt.Errorf("compile devcontainer.json: %w", err)
+					}
+					if buildParams.User != "" {
+						// BUG(mafredri): buildParams may set the user to remoteUser which
+						// is incorrect, we should only override if containerUser is set.
```
mafredri, line 322: Review: Didn't want to change this logic now, but documenting it for posterity.

envbuilder.go
```diff
@@ -385,7 +458,7 @@ func Run(ctx context.Context, opts options.Options) error {
 			// Since the user in the image is set to root, we also store the user
 			// in the magic file to be used by envbuilder when the image is run.
 			opts.Logger(log.LevelDebug, "writing magic image file at %q in build context %q", magicImageDest, magicTempDir)
-			if err := writeFile(opts.Filesystem, magicImageDest, 0o755, fmt.Sprintf("USER=%s\n", buildParams.User)); err != nil {
+			if err := writeMagicImageFile(opts.Filesystem, magicImageDest, runtimeData); err != nil {
```
mafredri, line 461: Review: Opted for a JSON encode instead of silly homebrew format.

envbuilder.go
```diff
@@ -556,23 +628,17 @@ func Run(ctx context.Context, opts options.Options) error {
 			return fmt.Errorf("restore mounts: %w", err)
 		}
 
-		// Create the magic file to indicate that this build
-		// has already been ran before!
-		file, err := opts.Filesystem.Create(magicDir.Built())
-		if err != nil {
-			return fmt.Errorf("create magic file: %w", err)
-		}
-		_ = file.Close()
```
mafredri, line 565: Review: Doing it later so it's applicable to both build and cached image. This might also fix an edge case (at least theoretically) where an error & restart could result in onCreateScript never being run.

envbuilder.go
```diff
@@ -586,117 +652,130 @@ func Run(ctx context.Context, opts options.Options) error {
 			}
 			opts.Logger(log.LevelInfo, "#%d: 👀 Found devcontainer.json label metadata in image...", stageNumber)
 			for _, container := range devContainer {
-				if container.RemoteUser != "" {
-					opts.Logger(log.LevelInfo, "#%d: 🧑 Updating the user to %q!", stageNumber, container.RemoteUser)
+				if container.ContainerUser != "" {
+					opts.Logger(log.LevelInfo, "#%d: 🧑 Updating the user to %q!", stageNumber, container.ContainerUser)
 
-					configFile.Config.User = container.RemoteUser
+					configFile.Config.User = container.ContainerUser
```
mafredri, line 658: Review: Decided to at least correct the behavior here. The remoteUser must not affect the user we run the container as, it's for devcontainer services & co.

envbuilder.go
```diff
@@ -586,117 +652,130 @@ func Run(ctx context.Context, opts options.Options) error {
...
+				maps.Copy(runtimeData.ContainerEnv, devContainer.ContainerEnv)
+				maps.Copy(runtimeData.RemoteEnv, devContainer.RemoteEnv)
+				if devContainer.ContainerUser != "" {
+					runtimeData.ContainerUser = devContainer.ContainerUser
 				}
-				os.Setenv(pair[0], pair[1])
+				runtimeData.Scripts = devContainer.LifecycleScripts
+			} else {
+				opts.Logger(log.LevelError, "Failed to parse devcontainer.json: %s", err.Error())
 			}
```
mafredri, line 704: Review: We're parsing the _workspace_ `devcontainer.json` here as we don't have access to the build-time `devcontainer.json`.

Motivation: Unless we cache the whole repo in the image, we can't reliably use `devcontainer.json` from it, it could be referencing unknown files from the repo which we haven't cached.

envbuilder.go
```diff
@@ -108,6 +159,30 @@ func Run(ctx context.Context, opts options.Options) error {
...
+		runtimeData.Image = true
+
+		// Some options are only applicable for builds.
+		if opts.RemoteRepoBuildMode {
+			opts.Logger(log.LevelDebug, "Ignoring %s option, it is not supported when using a pre-built image.", options.WithEnvPrefix("REMOTE_REPO_BUILD_MODE"))
+			opts.RemoteRepoBuildMode = false
+		}
+		if opts.ExportEnvFile != "" {
+			// Currently we can't support this as we don't have access to the
+			// post-build computed env vars to know which ones to export.
```
mafredri, line 179: Review: I couldn't think of a way to do this reliably. Maybe it's an easy fix but for now I'd like to leave it like this.

mafredri (commented):
I'll see about adding some tests tomorrow (need to figure out what to test), but in the meantime I think this is ready for review.

envbuilder.go
johnstcn: I think that's fair. We can address this in a follow-up.

envbuilder.go
johnstcn: Nice find. Can you open a follow-up PR to track this?

envbuilder.go
```diff
@@ -1311,14 +1407,14 @@ func execLifecycleScripts(
...
-	skippedRebuild bool,
+	firstStart bool,
 	userInfo userInfo,
 ) error {
 	if options.PostStartScriptPath != "" {
 		_ = os.Remove(options.PostStartScriptPath)
 	}
 
-	if !skippedRebuild {
+	if firstStart {
```
johnstcn, line 1417: 👍

johnstcn (approved):
I tested the provider against this branch and all tests passed 👍

envbuilder.go
```diff
@@ -58,32 +58,65 @@ var ErrNoFallbackImage = errors.New("no fallback image has been specified")
...
-		initArgs, err = shellquote.Split(opts.InitArgs)
-		if err != nil {
-			return fmt.Errorf("parse init args: %w", err)
-		}
+	err = syscall.Exec(args.InitCommand, append([]string{args.InitCommand}, args.InitArgs...), args.Environ)
+	if err != nil {
+		return fmt.Errorf("exec init script: %w", err)
 	}
 
+	return errors.New("exec failed")
```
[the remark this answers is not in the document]
mtojek: I admit I feel lost here... do we expect to always get an error from Run?

envbuilder.go
```diff
@@ -586,117 +652,130 @@ func Run(ctx context.Context, opts options.Options) error {
...
+		}
+	}
 
-			exportEnvFile.Close()
+	// Do not export env if we skipped a rebuild, because ENV directives
+	// from the Dockerfile would not have been processed and we'd miss these
+	// in the export. We should have generated a complete set of environment
+	// on the intial build, so exporting environment variables a second time
+	// isn't useful anyway.
+	if opts.ExportEnvFile != "" && !runtimeData.SkippedRebuild {
```
mtojek, line 752: nit: I believe this is a fundamental problem with this script. It would be cool to have `runtimeData` is immutable, and use helper functions instead `runtimeData.IsBuildSkipped()`

mtojek (approved):

envbuilder.go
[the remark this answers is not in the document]
johnstcn: If I'm not mistaken, doesn't syscall.Exec replace the current running process? If so, then nothing should be possible to happen after Exec()?

envbuilder.go
```diff
@@ -1241,6 +1319,24 @@ func RunCacheProbe(ctx context.Context, opts options.Options) (v1.Image, error)
 	return image, nil
 }
 
+func setEnvFromEtcEnvironment() error {
+	environ, err := os.ReadFile("/etc/environment")
+	if errors.Is(err, os.ErrNotExist) {
+		return nil
```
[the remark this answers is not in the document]
mafredri: It's not a requirement for this file to exist. So I think a log could be more confusing than not. Maybe debug level.

envbuilder.go
mafredri: I'm not sure how to achieve immutability here. For this particular function I suppose you could stat the file every time you want to use the bool, but seems like needless work to hit the fs each time. It doesn't work for most of the other data though.

envbuilder.go
```diff
@@ -175,112 +252,108 @@ func Run(ctx context.Context, opts options.Options) error {
...
+			file, err := opts.Filesystem.OpenFile(dockerfile, os.O_CREATE|os.O_WRONLY, 0o644)
 			if err != nil {
-				return fmt.Errorf("open devcontainer.json: %w", err)
+				return nil, err
 			}
 			defer file.Close()
-			content, err := io.ReadAll(file)
+			if opts.FallbackImage == "" {
+				if fallbackErr != nil {
+					return nil, xerrors.Errorf("%s: %w", fallbackErr.Error(), ErrNoFallbackImage)
```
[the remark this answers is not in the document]
mafredri: I avoided making any such changes to keep the diff smaller. I think unification is a great idea though, let's open a follow-up issue for it.

envbuilder.go
mafredri: https://github.com/coder/envbuilder/issues/364

envbuilder.go
mafredri: I did and closed it right up: https://github.com/coder/envbuilder/issues/365

I'll remove this comment.

There's still something wonky going on here: https://github.com/coder/envbuilder/blob/4f3e9cde3acf0ed45f3079293ba43be23889a47d/devcontainer/devcontainer.go#L301-L305

But I don't understand its purpose yet. It seems to change the user at the end without actually doing anything with that user?

envbuilder.go
mafredri: https://github.com/coder/envbuilder/issues/366

envbuilder.go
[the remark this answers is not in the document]
mafredri: https://github.com/coder/envbuilder/issues/367

johnstcn (approved):

options/options.go
```diff
@@ -573,4 +573,8 @@ func UnsetEnv() {
 		_ = os.Unsetenv(opt.Env)
 		_ = os.Unsetenv(strings.TrimPrefix(opt.Env, envPrefix))
 	}
+
+	// Unset the Kaniko environment variable which we set it in the
+	// Dockerfile to ensure correct behavior during building.
+	_ = os.Unsetenv("KANIKO_DIR")
```
mafredri, line 579: Review: This was left in a built image, but not in a cached image. I chose to go this path, but it also means it's not as "easy" to re-envbuilder inside a built image. But I didn't think that's _really_ a use-case, wdyt @johnstcn?

johnstcn: I think the main use-case of a cached image is to run the lifecycle scripts. We don't ever want to delete the filesystem or run kaniko inside a pre-built image anyway, so I don't see a reason to need this env.

mafredri: This isn't actually relevant for the cached image, since it doesn't set this env. It's only for a built image. But I'd argue the same logic applies there.

johnstcn (approved):
:shipit:

Issue #364, opened 2024-09-27
Defers run in `cmd/envbuilder` won't execute because Run performs `execve(2)`


mafredri:
Review: Even though we've fixed defers here, there's still `cmd/envbuilder` where e.g. `closeLogs` won't fire since we execve(2) after this.

_Originally posted by @mafredri in https://github.com/coder/envbuilder/pull/359#discussion_r1777477126_

johnstcn:
We could maybe add a `PreExec []func()` field to `envbuilder.Options` that will get called right before we `Exec`?

johnstcn:
Need to be careful about LogSenderV2 (see https://github.com/coder/envbuilder/pull/371); will fix this later.

PR #368, opened 2024-09-27
fix: add preExec to Run

johnstcn:
Fixes https://github.com/coder/envbuilder/issues/364

Adds the capability for `Run()` to run defers defined outside of `Run()`.
Currently the only use-case for this is to close the Coder logger.
Also adds an integration test that validates this behaviour and ensures that `closeLogs` gets called.

### DIFF
--- a/cmd/envbuilder/main.go
+++ b/cmd/envbuilder/main.go
@@ -37,6 +37,7 @@ func envbuilderCmd() serpent.Command {
 		Options: o.CLI(),
 		Handler: func(inv *serpent.Invocation) error {
 			o.SetDefaults()
+			preExec := make([]func(), 0)
 			o.Logger = log.New(os.Stderr, o.Verbose)
 			if o.CoderAgentURL != "" {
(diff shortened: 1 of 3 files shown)


cmd/envbuilder/main.go
```diff
@@ -37,6 +37,7 @@ func envbuilderCmd() serpent.Command {
 		Options: o.CLI(),
 		Handler: func(inv *serpent.Invocation) error {
 			o.SetDefaults()
+			preExec := make([]func(), 0)
```
mafredri, line 40: Let's make sure we never miss a cleanup!
```suggestion
			var preExec []func()
			defer func() { // Ensure cleanup in case of error.
				for _, fn := range preExec {
					fn()
				}
			}()
```
Also small nit: Slightly cleaner to prefer var or `:= []func(){}` assignment over `make(..., 0)`.

mafredri (approved):
LGTM!

PR #370, opened 2024-09-27
fix: use openfile to explicitly set magic image perms

mafredri:
This should fix an issue in our provider tests where build was producing an `rw-r--r--` and cache probing an `-rw-rw-r--` image file, resulting in a cache miss.


johnstcn (approved):
```
 TF_ACC=1 go test -count=1 ./internal/provider/ 
ok      github.com/coder/terraform-provider-envbuilder/internal/provider        37.437s
```

PR #375, opened 2024-09-30
fix: refactor coder logger to allow flush without deadlock

### LINKED ISSUE #364: Defers run in `cmd/envbuilder` won't execute because Run performs `execve(2)`

mafredri:
- **Reapply "fix: add preExec to Run (#368)" (#371)**
- **fix: refactor coder logger to allow flush without deadlock**

Fixes #364


mtojek (approved):
Refactor LGTM 👍

log/coder.go
```diff
@@ -95,65 +100,67 @@ func initRPC(ctx context.Context, client *agentsdk.Client, l slog.Logger) (proto
...
+			ls.Enqueue(uid, agentsdk.Log{
+				CreatedAt: time.Now(),
+				Output:    fmt.Sprintf(msg, args...),
+				Level:     codersdk.LogLevel(l),
+			})
+		}, func() {
+			closeOnce.Do(func() {
+				// Trigger a flush and wait for logs to be sent.
+				ls.Flush(uid)
+				ctx, cancel := context.WithTimeout(ctx, logSendGracePeriod)
```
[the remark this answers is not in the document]
mafredri: It's doable, but I feel their purpose is clearer as-is. We could actually make sendCtx not inherit function ctx instead.

log/coder.go
```diff
@@ -74,13 +78,14 @@ func initClient(coderURL *url.URL, token string) *agentsdk.Client {
...
 	var err error
-	retryCtx, retryCancel := context.WithTimeout(context.Background(), rpcConnectTimeout)
+	retryCtx, retryCancel := context.WithTimeout(ctx, rpcConnectTimeout)
 	defer retryCancel()
 	attempts := 0
 	for r := retry.New(100*time.Millisecond, time.Second); r.Wait(retryCtx); {
 		attempts++
 		// Maximize compatibility.
 		c, err = client.ConnectRPC20(ctx)
+		l.Info(ctx, "Connecting to Coder", slog.F("attempt", attempts), slog.F("error", err))
```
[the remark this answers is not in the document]
mafredri: Oh this was debugging I left in there 😅

cmd/envbuilder/main.go
```diff
@@ -50,6 +56,10 @@ func envbuilderCmd() serpent.Command {
 				if err == nil {
 					o.Logger = log.Wrap(o.Logger, coderLog)
 					defer closeLogs()
+					preExec = append(preExec, func() {
+						o.Logger(log.LevelInfo, "Closing logs")
+						closeLogs()
```
dannykopping, line 61: `closeLogs()` will be called twice when the function exits; is that intentional?

log/coder.go
```diff
@@ -44,18 +46,20 @@ func Coder(ctx context.Context, coderURL *url.URL, token string) (Func, func(),
 	}
 	if semver.Compare(semver.MajorMinor(bi.Version), minAgentAPIV2) < 0 {
 		metaLogger.Warn(ctx, "Detected Coder version incompatible with AgentAPI v2, falling back to deprecated API", slog.F("coder_version", bi.Version))
-		sendLogs, flushLogs := sendLogsV1(ctx, client, metaLogger.Named("send_logs_v1"))
-		return sendLogs, flushLogs, nil
+		logger, closer = sendLogsV1(ctx, client, metaLogger.Named("send_logs_v1"))
```
dannykopping, line 49: Nice 👍 I found those previous names hard to reason about.

log/coder.go
```diff
@@ -95,65 +99,67 @@ func initRPC(ctx context.Context, client *agentsdk.Client, l slog.Logger) (proto
...
+			if err := flushAndClose(ctx); err != nil {
 				l.Warn(ctx, "failed to flush logs", slog.Error(err))
 			}
 		}
 }
 
 // sendLogsV2 uses the v2 agent API to send logs. Only compatibile with coder versions >= 2.9.
-func sendLogsV2(ctx context.Context, dest agentsdk.LogDest, ls coderLogSender, l slog.Logger) (Func, func()) {
+func sendLogsV2(ctx context.Context, dest agentsdk.LogDest, ls coderLogSender, l slog.Logger) (logger Func, closer func()) {
+	sendCtx, sendCancel := context.WithCancel(ctx)
```
dannykopping, line 128: Should `sendCancel` be deferred, too?

log/coder_internal_test.go
```diff
@@ -221,6 +238,97 @@ func TestCoder(t *testing.T) {
 	})
 }
 
+//nolint:paralleltest // We need to replace a global timeout.
+func TestCoderRPCTimeout500MS(t *testing.T) {
```
dannykopping, line 242: Nit: do we have to include the timeout in the naming?

dannykopping (approved):

log/coder.go
mafredri: Nope, we want to keep this context alive until we've flushed or hit the timeout in the `closer` function.

log/coder_internal_test.go
mafredri: We don't, wanted to be explicit since we set it in the parent and use it in subtests, so that it's clear you shouldn't modify it without taking all tests into account. I'll leave a comment instead.

PR #377, opened 2024-09-30
reapply: "fix: refactor coder logger to allow flush without deadlock (#375)"

mafredri:
- **reapply: "fix: refactor coder logger to allow flush without deadlock (#375)"**
- **fix stackoverflow, paranoid teardown**

### DIFF
--- a/log/coder.go
+++ b/log/coder.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"sync"
 	"time"
 
 	"cdr.dev/slog"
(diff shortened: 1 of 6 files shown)


log/coder.go
```diff
@@ -44,18 +46,39 @@ func Coder(ctx context.Context, coderURL *url.URL, token string) (Func, func(),
...
 	dac, err := initRPC(ctx, client, metaLogger.Named("init_rpc"))
 	if err != nil {
 		// Logged externally
 		return nil, nil, fmt.Errorf("init coder rpc client: %w", err)
 	}
 	ls := agentsdk.NewLogSender(metaLogger.Named("coder_log_sender"))
 	metaLogger.Warn(ctx, "Sending logs via AgentAPI v2", slog.F("coder_version", bi.Version))
-	sendLogs, doneFunc := sendLogsV2(ctx, dac, ls, metaLogger.Named("send_logs_v2"))
-	return sendLogs, doneFunc, nil
+	logger, loggerCloser := sendLogsV2(ctx, dac, ls, metaLogger.Named("send_logs_v2"))
```
mafredri, line 69: Review: This used to be called closer, i.e. return variable name. After a last minute change in #375 we wrapped this function to ensure we can close drpcconn as well. This resulted in the variable being reassigned and calling itself.

```
 goroutine stack exceeds 1000000000-byte limit
runtime: sp=0xc048806380 stack=[0xc048806000, 0xc068806000]
fatal error: stack overflow

runtime stack:
runtime.throw({0x2088d94?, 0x200000001?})
	/usr/local/go/src/runtime/panic.go:1023 +0x5c fp=0x7ffdb0730768 sp=0x7ffdb0730738 pc=0x43e05c
runtime.newstack()
	/usr/local/go/src/runtime/stack.go:1103 +0x5bd fp=0x7ffdb0730918 sp=0x7ffdb0730768 pc=0x45a77d
runtime.morestack()
	/usr/local/go/src/runtime/asm_amd64.s:616 +0x7a fp=0x7ffdb0730920 sp=0x7ffdb0730918 pc=0x477b3a
```

log/coder.go
```diff
@@ -44,18 +46,39 @@ func Coder(ctx context.Context, coderURL *url.URL, token string) (Func, func(),
...
+		return logger, closer, nil
 	}
+
+	// Create a new context so we can ensure the connection is torn down.
+	ctx, cancel := context.WithCancel(ctx)
+	defer func() {
+		if err != nil {
+			cancel()
+		}
+	}()
```
mafredri, line 59: Review: This is a cautious addition to ensure there's no internal logic in `client.ConnectRPC20(ctx)` that might block or leave the connection open. This is to make it safer to call drpcconn close later.

mtojek (approved):
unblocking 👍

PR #378, opened 2024-09-30
fix: quote output of init command and args to prevent multiline log

mafredri:
Before:

```
=== Running the init command /usr/bin/echo [abra
cadabra
] as the "coder" user...
abra
cadabra
```

After:

```
=== Running init command as user "coder": ["/usr/bin/echo" "abra\ncadabra\n"]
abra
cadabra
```

Refs coder/coder#14882


johnstcn (approved):

dannykopping (approved):

mtojek (approved):
I know you can't go without my approval :)

PR #379, opened 2024-10-02
fix(scripts): always return the latest tag in version.sh

mafredri:
The `version.sh` script was prioritizing the first instead of last tag, which broke the v1.0.1 re-tag of v1.0.0.


dannykopping (approved):
LGTM, but perhaps this needs to be configurable?
Is the comment still valid?

mafredri:
> LGTM, but perhaps this needs to be configurable? Is the comment still valid?

For what purpose? The comment is perhaps a bit vague, but still valid IMO.

Issue #184, opened 2024-05-13
rename envbuilder.MagicDir to envbuilder.WorkingDir


johnstcn:
Not introduced by your PR but I want to point out that `MagicDir` is kinda unhelpful as a name; perhaps `EnvbuilderWorkingDir` would be more explicit?

_Originally posted by @dannykopping in https://github.com/coder/envbuilder/pull/183#discussion_r1598534110_

leo-oliveira2:
Hello, can I do this?

johnstcn:
> 1. Do we include a CLI change to rename the `MagicDirBase` option to `WorkingDirBase` in the scope for this issue? (please verify [@johnstcn](https://github.com/johnstcn)). I'm happy to do so. I just want the scope verified.

I would consider this in scope of this issue. It's not really a "CLI" change as such, as it's not an exposed CLI option. Only users who import `coder/envbuilder` would be affected by this.

> 3. I see the term magic used elsewhere in the code in relation to magic files, magic images and magic directives. Some of these come from the package I just renamed. Does `magic` have a specific definition that I'm unaware of? I fear that we might introduce confusion if the term magic somehow coupled these concepts.

It's just semantic bleed from `MagicDir`. If those can be renamed without any external effects, we could do so.

PR #388, opened 2024-10-15
chore(envbuilder.go): rename the concept of a magicDir to workingDir

SasSwart:
This PR is a first step towards #184. It does not close it entirely yet.
It renames the concept of a magicDir to "workingDir". There are no changes to this project's public API, nor to the CLI interface. Everything should therefore be backward compatible.

We do not yet rename the CLI option called [MagicDirBase](https://github.com/coder/envbuilder/blob/e448c03ae7bf7ab101416adc52848b15b17849f5/options/options.go#L168), because this could potentially break things for consumers of the CLI, such as the Terraform provider. This will be left to a subsequent PR.


johnstcn (approved):
All of these changes are internal-only and don't seem to pose any external risk 👍

PR #389, opened 2024-10-15
fix(envbuilder.go)!: rename MagicDirBase CLI option

### LINKED ISSUE #184: rename envbuilder.MagicDir to envbuilder.WorkingDir

SasSwart:
This PR is a breaking change!
It renames a publicly accessible CLI option referenced in #184.
It should not be merged into release/1.0 as this would break at least the terraform provider that relies on this option.

This commit may be cherry-picked into release/2.0 when that branch is created.

This PR closes #184.

### DIFF
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -120,7 +120,7 @@ func Run(ctx context.Context, opts options.Options, preExec ...func()) error {
 func run(ctx context.Context, opts options.Options, execArgs *execArgsInfo) error {
 	defer options.UnsetEnv()
 
-	workingDir := workingdir.At(opts.MagicDirBase)
+	workingDir := workingdir.At(opts.WorkingDirBase)
 
 	stageNumber := 0
 	startStage := func(format string, args ...any) func(format string, args ...any) {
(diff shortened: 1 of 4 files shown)


johnstcn:
@mafredri Are you happy with us merging this now? Or would you like to defer this to a later date?

mafredri:
> @mafredri Are you happy with us merging this to `main` now? Or would you like to defer this to a later date so we don't end up needing to do too many backport changes?

(We could just make it backwards compatible for a period of time by including the old property (and checking it too) but marking it deprecated. That way we’re not in a rush to sync the changes to the provider.)

I’m also fine with merging as-is, but I’d avoid cherry-picking it into release/1.0 unless there are conflicts. Our next non-patch release should be release/1.1, which we branch out from `main`, and I think it’s fine to include this there.

mafredri (approved):

Issue #93, opened 2024-02-28
envbuilder support for build-time secrets is undocumented


ns-mkusper:
There doesn't seem to be any way to pass build-time secrets when using envbuilder, leaving me with no straight-forward option to include any Dockerfile with commands similar to:

`RUN --mount=type=secret,id=test-token poetry config http-basic.test-token-pypi gitlab-ci-token $(cat /run/secrets/test-token)`

I'd be willing to drop a patch for this, but I'm curious to hear opinions on whether or not this is something envbuilder can support without violating its design principles.

ns-mkusper:
Seems that the standard kaniko method of mounting secrets in `/kaniko` does work to resolve this issue. This works for me and I'm willing to use it permanently but seems like something of a workaround.

SasSwart:
Hey There!

Support for build secrets has been formalised and added as of #401.
Documentation is available here:
https://github.com/coder/envbuilder/blob/main/docs/build-secrets.md

This feature is as yet unreleased. Keep an eye on our releases and changelog for further information.
If you need this feature urgently, feel free to build your own envbuilder container using this script:
https://github.com/coder/envbuilder/blob/main/scripts/build.sh like this:
```
./scripts/build.sh --base=test-envbuilder
```

SasSwart:
Builds secrets are now also supported in the Terraform provider for envbuilder.
This was the last feature we needed to close this issue.

Closing.

Issue #366, opened 2024-09-27
Figure out if we can support ExportEnvFile when running from a cached image


mafredri:
Review: I couldn't think of a way to do this reliably. Maybe it's an easy fix but for now I'd like to leave it like this.

_Originally posted by @mafredri in https://github.com/coder/envbuilder/pull/359#discussion_r1777515547_

PR #391, opened 2024-10-16
feat(envbuilder.go): add support for build secrets

SasSwart:
This PR adopts support for build secrets from our recent updates to Kaniko to provide a safer alternative to this: https://github.com/coder/envbuilder/issues/93


johnstcn (commented):
This is somewhat different from what I had in mind, and seems to cross more into the direction of https://github.com/coder/envbuilder/issues/366

The idea is that we want these secrets to be available during the container build process only. As far as I can tell, this persists them to disk inside Kaniko's working directoryl.

What do you think about the following approach:
- Fetch all environment variables with the prefix `ENVBUILDER_SECRET` and store them in a map or similar
- Before building the image using Kaniko, set all the environment variables from above without `ENVBUILDER_SECRET` prefix.
- After building the image and just before exec'ing the init command, unset all of the above environment variables.

If the prefix `ENVBUILDER_SECRET_` is misleading, we can rename it to something else also.

SasSwart:
> Before building the image using Kaniko, set all the environment variables from above without ENVBUILDER_SECRET prefix.

I did attempt this, but wasn't able to get it to work. each RUN directive starts with a mostly blank environment as far as I can tell. I will revisit this approach though. Perhaps I can get it to work upon closer inspection.

Out of interest, why do you prefer the ENV approach as opposed to persisting them to disk?

johnstcn:
> Out of interest, why do you prefer the ENV approach as opposed to persisting them to disk?

I actually don't have a good answer to this apart from 'environment variables are easy and it seems to be the method of least surprise'.

However, I can see some folks being wary of passing build-time secrets via environment variable, as someone with access to either the container spec or some 'back-door' into the container (e.g. via `kubectl exec`) would then potentially end up being able to read the env var.

SasSwart:
@johnstcn The updated PR should demonstrate a more flexible approach in line with docker standards. The user can now choose whether they want an ENV, or a file. They can also specify where that file should be.

envbuilder.go
```diff
@@ -525,10 +525,15 @@ func run(ctx context.Context, opts options.Options, execArgs *execArgsInfo) erro
 			if val, ok := os.LookupEnv("KANIKO_REGISTRY_MIRROR"); ok {
 				registryMirror = strings.Split(val, ";")
 			}
-			var destinations []string
+			var destinations = []string{"image"}
```
SasSwart, line 528: Note to self: undo this

envbuilder.go
```diff
@@ -538,6 +543,7 @@ func run(ctx context.Context, opts options.Options, execArgs *execArgsInfo) erro
 				RunStderr:          stderrWriter,
 				Destinations:       destinations,
 				NoPush:             !opts.PushImage || len(destinations) == 0,
+				TarPath:            filepath.Join(magicTempDir.Path(), "image.tar"),
```
SasSwart, line 546: Note to self: undo this

envbuilder.go
```diff
@@ -579,7 +586,7 @@ func run(ctx context.Context, opts options.Options, execArgs *execArgsInfo) erro
 				return nil, xerrors.Errorf("do build: %w", err)
 			}
 			endStage("🏗️ Built image!")
-			if opts.PushImage {
+			if opts.PushImage || true {
```
SasSwart, line 589: Note to self: undo this

SasSwart (commented):
Self Review:

mafredri (commented):
In general the implementation looks good, had some minor suggestions. Tests need a bit more work though.

options/secrets_test.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+func TestGetBuildSecrets(t *testing.T) {
+	t.Parallel()
+
+	t.Run("no secrets", func(t *testing.T) {
+		t.Parallel()
+
+		// given
+		os.Clearenv()
```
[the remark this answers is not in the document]
SasSwart: 🤦🏼 You're absolutely right. Fixing.

integration/integration_test.go
```diff
@@ -897,6 +896,69 @@ func TestUnsetOptionsEnv(t *testing.T) {
...
+	srv := gittest.CreateGitServer(t, gittest.Options{
+		Files: map[string]string{
+			".devcontainer/devcontainer.json": `{
+				"name": "Test",
+				"build": {
+					"dockerfile": "Dockerfile"
+				},
+			}`,
+			".devcontainer/Dockerfile": "FROM " + testImageAlpine +
+				"\nRUN --mount=type=secret,id=FOO cat /run/secrets/FOO > /foo_from_file" +
```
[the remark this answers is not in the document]
SasSwart: Test cases added.

SasSwart:
@mafredri I've tended to all review notes. Thanks for all the feedback, it was useful!
Tests should be in a much better space now. Please have another look?

options/secrets.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"fmt"
+	"os"
+
+	"github.com/coder/serpent"
+)
+
+var buildSecretPrefix = fmt.Sprintf("%sBUILD_SECRET_", envPrefix)
+
+func EnvWithBuildSecretPrefix(secretName, secretValue string) string {
```
mafredri, line 12: Suggestion: It's best practice for exported functions to have documentation (`// EnvWithBuildSecretPrefix ...`). It's part of the original Go linter (https://pkg.go.dev/github.com/golang/lint) and effective go, so a bit surprising.

```
❯ golint ./...
envbuilder.go:43:2: a blank import should be only in a main or test package, or have a comment justifying it
devcontainer/devcontainer.go:34:6: exported type Spec should have comment or be unexported
devcontainer/devcontainer.go:50:6: exported type LifecycleScripts should have comment or be unexported
devcontainer/devcontainer.go:57:6: exported type BuildSpec should have comment or be unexported
devcontainer/devcontainer.go:78:1: exported function SubstituteVars should have comment or be unexported
devcontainer/script.go:16:6: exported type LifecycleScript should have comment or be unexported
devcontainer/script.go:21:1: exported method LifecycleScript.IsEmpty should have comment or be unexported
```

It seems `golint` is marked deprecated in `golangci-lint` though, so I guess the alternative linters aren't enforcing this rule. 😔 I always found it slightly annoying but ultimately worth the annoyance 😄.

options/secrets_test.go
```diff
@@ -0,0 +1,150 @@
...
+				"ENVBUILDER_BUILD_SECRET_BAZ": "qux",
+			},
+			expectedSecrets: []string{"FOO=bar", "BAZ=qux"},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			envSnapshot := snapshotEnv()
+			t.Cleanup(func() { assertEnvUnchanged(t, envSnapshot) })
```
mafredri, line 45: This could lead to flakiness if someone is running these tests with a secret environment variable set in their shell env since we're using `os.Unsetenv` below and never restoring it.

options/secrets_test.go
```diff
@@ -0,0 +1,150 @@
...
+		})
+	}
+}
+
+func TestClearBuildSecrets(t *testing.T) {
+	tests := []struct {
+		name                       string
+		initialEnvVars             map[string]string
+		expectedSecretsBeforeClear []string
+		expectedSecretsAfterClear  []string
```
mafredri, line 69: Never used, bug or consider removing?

options/secrets_test.go
```diff
@@ -0,0 +1,150 @@
...
+		currentValue, exists := os.LookupEnv(key)
+		assert.True(t, exists, "expected environment variable %s to be set", key)
+		assert.Equal(t, expectedValue, currentValue, "expected environment variable %s to be unchanged", key)
+	}
+	for _, envVar := range os.Environ() {
+		key := strings.SplitN(envVar, "=", 2)[0]
+		if _, exists := snapshot[key]; !exists {
+			assert.Fail(t, "unexpected environment variable set", "variable %s was set during the test", key)
+		}
+	}
```
mafredri, line 149: Could perhaps be simplified?
```go
got := snapshotEnv()
require.Equal(t, snapshot, got)
```

options/secrets_test.go
```diff
@@ -0,0 +1,150 @@
...
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/coder/envbuilder/options"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetBuildSecrets(t *testing.T) {
```
mafredri, line 13: Suggestion: In coder/coder we usually add comments like this:
```go
//nolint:tparallel // This test modifies the environment.
```
It seems we don't have this linter enabled here, but it's good practice to add a note about why we're not doing a parallel test.

mafredri (approved):
A few minor nits but otherwise looking good. 👍🏻

envbuilder.go
```diff
@@ -525,10 +525,15 @@ func run(ctx context.Context, opts options.Options, execArgs *execArgsInfo) erro
 			if val, ok := os.LookupEnv("KANIKO_REGISTRY_MIRROR"); ok {
 				registryMirror = strings.Split(val, ";")
 			}
-			var destinations []string
+			destinations := []string{}
```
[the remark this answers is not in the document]
SasSwart: I made this change because I was requested to do so by a linter, but perhaps I could have made it in a way that is not semantically different. Will check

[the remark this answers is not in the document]
mafredri: Which linter? `make lint` doesn't trigger this for me 🤔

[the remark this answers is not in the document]
SasSwart: It was a linter in CI. I'll find the failure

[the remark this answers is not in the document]
SasSwart: Ah! This makes sense now:
https://github.com/coder/envbuilder/actions/runs/11495941565/job/31996550118

I was forcing kaniko to output a tarball by adding a default item to this.
Without the default item, this can revert back to its original declaration.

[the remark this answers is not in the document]
SasSwart: Fixed

options/secrets_test.go
SasSwart: I'd prefer to enable the linter before adding the comment, because that way we can confirm that it works. I've created a separate issue for updating the linter config though, because if it finds more than expected I don't want that to extend this PR. I've created #399 to deal with linting for envbuilder, including this comment and the other one in this PR.

options/secrets.go
SasSwart: I've documented the function and created an issue for us to support these linting cases in envbuilder.

Issue #333, opened 2024-09-01
Allow passing Git SSH key as an environment variable


phorcys420:
Following up to a [discord conversation](https://discord.com/channels/747933592273027093/971231372373033030/1279228570295599114) where an user tries to clone a Git repository via SSH and is met with the following error:
```
envbuilder v1.0.0-rc.6+23d086e - Build development environments from repositories in a container
coder/coder#1: 🔑 Using SSH authentication!
coder/coder#1: 🔑 No SSH key found, falling back to agent!
coder/coder#1: ❌ Failed to connect to SSH agent: error creating SSH agent: "SSH agent requested but SSH_AUTH_SOCK not-specified"
coder/coder#1: 📦 Cloning git@github.com:XXX.git to /workspaces/XXX...
Failed to clone repository: clone "git@XXX.git": error creating SSH agent: "SSH agent requested but SSH_AUTH_SOCK not-specified"

Falling back to the default image...
```

This makes sense because the Coder user's SSH key was never added to the workspace.

While there is a workaround, it's really not great, how can we make this better?
With k8s the issue is not that bad since you can just define a secret from the key and pass that secret as a file to the container, but what about Docker and VMs?

## Workaround
[Downloading the key](https://github.com/coder/modules/blob/main/git-commit-signing%2Frun.sh#L18-L36) to some temporary path (via an entrypoint script, since envbox runs before the agent)
(or via the TF provider's [`coder_workspace_owner.ssh_*`](https://registry.terraform.io/providers/coder/coder/latest/docs/data-sources/workspace_owner#read-only) instead of using a script)
Then passing the key's path to `ENVBUILDER_GIT_SSH_PRIVATE_KEY_PATH`.

2o1o0:
Hey,
Not sure I'm using the right issue or if I should create a new one, but I ended up here after following [this](https://github.com/coder/envbuilder/pull/174) and [this](https://github.com/coder/terraform-provider-coder/issues/219).

So I'm trying to spin up some workspaces in K8S using some .devcontainers on a remote public gitlab repository that require authentication. As of now, I'd like to use the ssh key provided to the user within his profile to access this repository. In my case my Coder instance is setup to use OIDC from a local Keycloak instance.

My understanding is that envbuilder [can do this](https://github.com/coder/envbuilder/blob/main/docs/git-auth.md), if you provide a path to private key file or use SSH_AUTH_SOCK. And I'm suppose to use the first option in a Kubernetes context.
[...]

PR #396, opened 2024-10-25
feat: add env var for ssh private key

DanielleMaywood:
Closes https://github.com/coder/envbuilder/issues/333

Allow passing a base64 encoded ssh private key to `ENVBUILDER_GIT_SSH_PRIVATE_KEY_BASE64` and using this for git authentication.

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -2,6 +2,7 @@ package git
 
 import (
 	"context"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"io"
@@ -181,6 +182,22 @@ func ReadPrivateKey(path string) (gossh.Signer, error) {
 	return k, nil
 }
 
+// DecodeBase64PrivateKey attempts to decode a base64 encoded private
+// key and returns an ssh.Signer
+func DecodeBase64PrivateKey(key string) (gossh.Signer, error) {
+	bs, err := base64.StdEncoding.DecodeString(key)
+	if err != nil {
+		return nil, fmt.Errorf("decode private key: %w", err)
+	}
+
+	k, err := gossh.ParsePrivateKey(bs)
+	if err != nil {
+		return nil, fmt.Errorf("parse private key: %w", err)
+	}
+
+	return k, nil
+}
+
 // LogHostKeyCallback is a HostKeyCallback that just logs host keys
 // and does nothing else.
 func LogHostKeyCallback(logger func(string, ...any)) gossh.HostKeyCallback {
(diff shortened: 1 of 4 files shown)


git/git.go
```diff
@@ -273,6 +290,17 @@ func SetupRepoAuth(logf func(string, ...any), options *options.Options) transpor
 		}
 	}
 
+	// If no path was provided, fall back to the environment variable
+	if options.GitSSHPrivateKeyBase64 != "" {
```
mafredri, line 294: It might be worth thinking about what the behavior should be if _both_ key path and key base64 is given. Do we exit immediately (invalid input)? Do we prefer one over the other? How do we inform the user?

options/options.go
```diff
@@ -363,6 +366,12 @@ func (o *Options) CLI() serpent.OptionSet {
 			Value:       serpent.StringOf(&o.GitSSHPrivateKeyPath),
 			Description: "Path to an SSH private key to be used for Git authentication.",
 		},
+		{
+			Flag:        "git-ssh-private-key-base64",
+			Env:         WithEnvPrefix("GIT_SSH_PRIVATE_KEY_BASE64"),
+			Value:       serpent.StringOf(&o.GitSSHPrivateKeyBase64),
+			Description: "SSH private key to be used for Git authentication.",
```
mafredri, line 373:
```suggestion
			Description: "Base64 encoded SSH private key to be used for Git authentication. This option takes precedence over the private key path option.",
```
The "This option takes ..." is just there for completeness, may be rewritten depending on how the final implementation turns out.

mafredri (commented):
I think we should also have an integration test to confirm that this works as expected. Perhaps also validating the behavior when both file and base64 is given.

git/git.go
johnstcn: To be honest, I think we should error if both are provided. Removes any ambiguity.

johnstcn:
> I think we should also have an integration test to confirm that this works as expected. Perhaps also validating the behavior when both file and base64 is given.

There's a test SSH server implementation you can probably re-use here: https://github.com/coder/terraform-provider-envbuilder/blob/main/internal/provider/git_test.go#L85

options/options.go
johnstcn: Yep, we should be explicit that both options are mutually exclusive.

mafredri:
> There's a test SSH server implementation you can probably re-use here: [coder/terraform-provider-envbuilder@`main`/internal/provider/git_test.go#L85](https://github.com/coder/terraform-provider-envbuilder/blob/main/internal/provider/git_test.go?rgh-link-date=2024-10-25T09%3A58%3A25Z#L85)

Yep, that should work. I believe only the `PublicKeyHandler` implementation needs to be updated to reject/accept the key.

options/options.go
DanielleMaywood: Have documented this now 👍

johnstcn (commented):
Once we have an integration test, this will be good to go! 👍

DanielleMaywood:
I've added an integration test. It works similar to the existing tests in `git_test.go`. Rather than checking it has worked without error (as the current git ssh server impl doesn't appear to work properly), it instead checks that the error is one that comes _after_ auth has ran. https://github.com/coder/envbuilder/blob/58ac15f45d3d3ee8479d769cc7c3dfcc7ca64f87/git/git_test.go#L266-L268

murrayju:
I've found that this works, but I couldn't find an example anywhere, so it took a very long time to figure out. Sharing in case this helps someone like me in the future:

```tf
locals {
  # ...
  envbuilder_env = {
    # ...
    "ENVBUILDER_GIT_SSH_PRIVATE_KEY_BASE64" : base64encode(data.coder_workspace_owner.me.ssh_private_key),
  }
  # Convert the above map to the format expected by the docker provider.
}
```

johnstcn:
Thanks for the heads-up @murrayju ! I've added a PR to document this feature.

https://github.com/coder/envbuilder/pull/442

murrayju:
Thanks, that looks helpful. In particular, I was using [this coder template](https://github.com/coder/coder/blob/main/examples/templates/docker-devcontainer/main.tf#L130). I'd suggest that this template should be updated to include this config, so that private repos can be used.

PR #397, opened 2024-10-28
chore(integration): update TestLogs to validate that CODER_ASGENT_SUBSYSTEM is set correctly

johnstcn:
We had no test coverage that `CODER_AGENT_SUBSYSTEM` was actually being set. This adds a check to ensure that it is.

### DIFF
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -123,10 +123,11 @@ func TestLogs(t *testing.T) {
 			"Dockerfile": fmt.Sprintf(`FROM %s`, testImageUbuntu),
 		},
 	})
-	_, err := runEnvbuilder(t, runOpts{env: []string{
+	ctrID, err := runEnvbuilder(t, runOpts{env: []string{
 		envbuilderEnv("GIT_URL", srv.URL),
 		"CODER_AGENT_URL=" + logSrv.URL,
 		"CODER_AGENT_TOKEN=" + token,
+		"ENVBUILDER_INIT_SCRIPT=env",
 	}})
 	require.NoError(t, err)
 	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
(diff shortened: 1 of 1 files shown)


integration/integration_test.go
```diff
@@ -136,6 +137,28 @@ func TestLogs(t *testing.T) {
...
+	// Wait for the container to exit
+	client, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
+	require.NoError(t, err)
+	require.Eventually(t, func() bool {
+		status, err := client.ContainerInspect(ctx, ctrID)
+		if !assert.NoError(t, err) {
+			return false
+		}
+		return !status.State.Running
+	}, 10*time.Second, time.Second, "container never exited")
```
mafredri, line 150: Could 10 seconds be potentially flaky?

mafredri (approved):

integration/integration_test.go
```diff
@@ -136,6 +137,28 @@ func TestLogs(t *testing.T) {
...
+	// Check the expected log output
+	logReader, err := client.ContainerLogs(ctx, ctrID, container.LogsOptions{
+		ShowStdout: true,
+		ShowStderr: true,
+	})
+	require.NoError(t, err)
+	logBytes, err := io.ReadAll(logReader)
+	require.NoError(t, err)
+	logs := string(logBytes)
+	require.Contains(t, logs, "CODER_AGENT_SUBSYSTEM=envbuilder")
```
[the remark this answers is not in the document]
johnstcn: It only gets set if `CODER_AGENT_URL` and `CODER_AGENT_TOKEN` are set. It's not set otherwise. This is more of a convenient place to co-locate it.

integration/integration_test.go
johnstcn: Potentially, but I think it's relatively unlikely that it will take that long to execute `env` and for the container to exit.

PR #398, opened 2024-10-28
fix: search `$DOCKER_CONFIG` if no base64 config is provided

### LINKED ISSUE #392: Allow mounting in DOCKER_CONFIG again
# Motivation

In https://github.com/coder/envbuilder/pull/336 we temporarily override `DOCKER_CONFIG` to set the config specified by `ENVBUILDER_DOCKER_CONFIG_BASE64`. 

This breaks some existing users who were mounting in a docker config and specifying the `DOCKER_CONFIG` environment variable in their image.

# Proposed solution

Only override `DOCKER_CONFIG` if `ENVBUILDER_DOCKER_CONFIG_BASE64` is set. Otherwise, leave it as-is.

mafredri:
Fixes #392


envbuilder.go
```diff
@@ -1627,55 +1627,83 @@ func parseMagicImageFile(fs billy.Filesystem, path string, v any) error {
...
+		}
+		logf(log.LevelInfo, "Set DOCKER_CONFIG to %s", newDockerConfig)
+
+		restoreEnv = func() error {
+			// Restore the old DOCKER_CONFIG value.
+			if oldDockerConfig == "" {
+				err := os.Unsetenv("DOCKER_CONFIG")
+				if err != nil {
+					err = fmt.Errorf("unset DOCKER_CONFIG: %w", err)
+				}
```
johnstcn, line 1651: suggestion: could this also be simplified?

```
return fmt.Errorf(...)
```

integration/integration_test.go
```diff
@@ -2357,6 +2412,9 @@ func runEnvbuilder(t *testing.T, opts runOpts) (string, error) {
 	logChan, errChan := streamContainerLogs(t, cli, ctr.ID)
 	go func() {
 		for log := range logChan {
+			if opts.logbuf != nil {
+				opts.logbuf.WriteString(log)
+			}
```
johnstcn, line 2417: nice 👍  I'll have to go through and use that in other places!

johnstcn (approved):

envbuilder.go
mafredri: Surely! Had a log before, forgot to simplify.

integration/integration_test.go
mafredri: I forgot we had another way to get the logs 😂, but alas this seems more convenient.

mafredri:
CI seems to have issue with privilege and remounting, not sure why it applies to my test and not others. Looking into it.

Also realized we'll have to copy over `config.json` or change `DOCKER_CONFIG` to the remounted path if applicable.

envbuilder.go
```diff
@@ -711,6 +711,11 @@ func run(ctx context.Context, opts options.Options, execArgs *execArgsInfo) erro
 	// Sanitize the environment of any opts!
 	options.UnsetEnv()
 
+	// Remove the Docker config secret file!
+	if err := cleanupDockerConfigOverride(); err != nil {
+		return err
+	}
```
mafredri, line 717: Review: Moved this higher up so that when we reset DOCKER_CONFIG, we don't interfere with envbuilder trying to set envs from the build/devcontainer.

integration/integration_test.go
```diff
@@ -552,13 +552,21 @@ func TestBuildFromDockerfile(t *testing.T) {
...
+			envbuilderEnv("DOCKERFILE_PATH", "Dockerfile"),
+			envbuilderEnv("DOCKER_CONFIG_BASE64", base64.StdEncoding.EncodeToString([]byte(`{"experimental": "enabled"}`))),
+			"DOCKER_CONFIG=/config", // Ignored, because we're setting DOCKER_CONFIG_BASE64.
+		},
+		logbuf: logbuf,
+	})
 	require.NoError(t, err)
 
+	require.Contains(t, logbuf.String(), "Set DOCKER_CONFIG to /.envbuilder")
+	require.NotContains(t, logbuf.String(), "Using DOCKER_CONFIG at ")
```
SasSwart, line 568: Do we generally trust log output for test assertions?

If this log line ever drifts, this test will become a silent false positive. Is there a practical way to assert against the actual behaviour of the system? Could we copy the written docker config secret file so that it makes its way into the container and then assert against its contents?

Would that approach be worth it?

johnstcn: We could modify `ENVBUILDER_INIT_SCRIPT` to write the content of `$DOCKER_CONFIG`

mafredri: @SasSwart good points, I agree, in general it’s best to test behavior and not output. @johnstcn I wanted to do that here but we can’t do it in the init script since it runs after cleanup.

Not sure if it’ll work but we could try to copy it as part of Dockerfile build instructions, will look into that.

mafredri: Went back to the drawing board with the implementation, changed the behavior a bit for user-provided config files and converted the integration tests to primarily testing behavior.

envbuilder.go
```diff
@@ -1627,55 +1657,161 @@ func parseMagicImageFile(fs billy.Filesystem, path string, v any) error {
...
+	for _, path := range []string{
+		oldDockerConfigFile,               // $DOCKER_CONFIG/config.json
+		oldDockerConfig,                   // $DOCKER_CONFIG
+		workingDir.Join(dockerConfigFile), // /.envbuilder/config.json
+	} {
+		if path == "" || !fileExists(bfs, path) {
+			continue
+		}
+
+		logf(log.LevelWarn, "Found Docker config at %s, this file will remain after the build", path)
```
johnstcn, line 1705: :+1:

integration/integration_test.go
```diff
@@ -545,27 +545,142 @@ func TestBuildFromDevcontainerWithFeatures(t *testing.T) {
...
-	configJSONContainerPath := workingdir.Default.Join("config.json")
-	output = execContainer(t, ctr, "stat "+configJSONContainerPath)
-	require.Contains(t, output, "No such file or directory")
+			tt.validate(t, tt, ctrID, logbuf.String())
+
+			// Always verify that the Docker configuration secret file is removed.
+			output := execContainer(t, ctrID, "stat "+configJSONContainerPath)
+			require.Contains(t, output, "No such file or directory")
+		})
+	}
```
johnstcn, line 683: 👍 👍

integration/integration_test.go
```diff
@@ -545,27 +545,142 @@ func TestBuildFromDevcontainerWithFeatures(t *testing.T) {
...
+		want := tc.configBase64
+		if want == "" {
+			want = tc.configFile.data
+		}
+		if want != "" {
+			require.Contains(t, logs, "Set DOCKER_CONFIG to /.envbuilder/.docker")
+			require.Equal(t, want, got)
+		}
+
+		// Ensure that a warning message is printed if config secrets
```
SasSwart, line 579: Why do we only warn on this? Users may have a reason to use a DOCKER_CONFIG at build time and they may have a reason to use one at run time. But those are separate use cases as far as I can tell. Should we allow the line between the two to blur?

Would it be more secure and user friendly to treat build time DOCKER_CONFIG like a build secret; ensuring that it's always cleared?

Runtime secrets could then be provided using a runtime mechanism? Right users are mounting it in, but to me this would be treated just like any other runtime secret.

johnstcn (approved):
I like this approach of warning the user.
If possible, I'd also like to see the documentation around registry auth updated either as part of this PR or as a follow-up.

integration/integration_test.go
mafredri: > Why do we only warn on this? Users may have a reason to use a DOCKER_CONFIG at build time and they may have a reason to use one at run time. But those are separate use cases as far as I can tell. Should we allow the line between the two to blur?

That's a valid question. We can't determine the user's intent here. I'd say it's more likely that a given `DOCKER_CONFIG` is intended for build time since there's no guarantee Docker will be available in the resulting devcontainer.

Adding to this, if the devcontainer should have a `DOCKER_CONFIG` set at runtime, that should be done via `containerEnv` in `devcontainer.json`, not on the envbuilder container.

> Would it be more secure and user friendly to treat build time DOCKER_CONFIG like a build secret; ensuring that it's always cleared?

Perhaps, but this ties into why we "only warn". We can't guarantee that we'll be able to wipe the secrets if the user has mounted them in, possibly as read-only, and the container doesn't have unmount permissions (or, will remounting be possible if there's sudo?). That's why we only warn and don't attempt to clear. If we attempted something else, we may risk introducing behavior that varies depending on the environment it's run in. It's better if the user knows what to expect.

Also, since we restore the original `DOCKER_CONFIG`, we're allowing that config location to be used at runtime, enabling use-cases where you have a separate build and runtime config. You set the build config via base64 env (takes precedence), and after the build, the location will be used at runtime. As a side-effect, the same config can be used for _both build and runtime_, which might not be ideal, but what's the alternative?

> Runtime secrets could then be provided using a runtime mechanism? Right users are mounting it in, but to me this would be treated just like any other runtime secret.

Not sure what you're referring to by runtime secret and runtime mechanism? Could you expand on this?

PR #400, opened 2024-10-29
test: enable parallel on more integration tests

mafredri:


SasSwart (approved):
I don't see any obvious hints that this could introduce flakes. LGTM

johnstcn (approved):

PR #408, opened 2024-11-06
chore(docs): add explicit documention for parts of devcontainer spec supported by envbuilder

johnstcn:


docs/devcontainer-spec-support.md
```diff
@@ -0,0 +1,72 @@
...
+| 🟢                            | `containerEnv`                        | Environment variables to set inside the container.                                                                     | -                                                      |
+| 🟢                            | `remoteEnv`                           | Override environment variables for specific supporting services or tools, but not the container as a whole.            | -                                                      |
+| 🟢                            | `remoteUser`                          | Override the user that supporting tools run as inside the container without changing the user for the whole container. | -                                                      |
+| 🟢                            | `containerUser`                       | Override the user for all operations run inside the container.                                                         | -                                                      |
```
mafredri, line 24: With these, we may have some divergence from the spec. I think we can leave it green but we may need to verify/disclaimer.

docs/devcontainer-spec-support.md
```diff
@@ -0,0 +1,72 @@
...
+| 🟢                            | `features`                            | Features to be added to the devcontainer.                                                                              | -                                                      |
+| 🔴                            | `overrideFeatureInstallOrder`         | Override the order in which features should be installed.                                                              | [#226](https://github.com/coder/envbuilder/issues/226) |
+| 🔴                            | `customizations`                      | Product-specific properties, e.g., _VS Code_ settings and extensions.                                                  | [#43](https://github.com/coder/envbuilder/issues/43)   |
+| **Image/Dockerfile**          |                                       |                                                                                                                        |
```
mafredri, line 37: Might be clearer to split these into separate tables instead of "sections".

docs/devcontainer-spec-support.md
```diff
@@ -0,0 +1,72 @@
...
+| 🔴                            | `mounts`                              | Add additional mounts to the container.                                                                                | [#220](https://github.com/coder/envbuilder/issues/220) |
+| 🟢                            | `features`                            | Features to be added to the devcontainer.                                                                              | -                                                      |
+| 🔴                            | `overrideFeatureInstallOrder`         | Override the order in which features should be installed.                                                              | [#226](https://github.com/coder/envbuilder/issues/226) |
+| 🔴                            | `customizations`                      | Product-specific properties, e.g., _VS Code_ settings and extensions.                                                  | [#43](https://github.com/coder/envbuilder/issues/43)   |
```
mafredri, line 36: I'm not sure red/unsupported is right here. I see this more as a responsibility of the tool that's being customized to apply them to itself.

Maybe a new category like gray = not envbuilders responsibility. 😅

mafredri (approved):
Nice work!

docs/devcontainer-spec-support.md
johnstcn: Can do!

docs/devcontainer-spec-support.md
johnstcn: 🤔 If it's not envbuilder's responsibility, why is there an open issue to support VSCode-specific customizations?

docs/devcontainer-spec-support.md
```diff
@@ -0,0 +1,72 @@
...
+| 🔴                            | `capAdd`                              | Capabilities to add to the container (for example, `SYS_PTRACE`).                                                      | -                                                      |
+| 🔴                            | `securityOpt`                         | Security options to add to the container (for example, `seccomp=unconfined`).                                          | -                                                      |
+| 🔴                            | `mounts`                              | Add additional mounts to the container.                                                                                | [#220](https://github.com/coder/envbuilder/issues/220) |
+| 🟢                            | `features`                            | Features to be added to the devcontainer.                                                                              | -                                                      |
```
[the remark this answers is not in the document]
johnstcn: Good catch1

docs/devcontainer-spec-support.md
```diff
@@ -0,0 +1,72 @@
...
+The symbols in the first column indicate the support status:
+
+- 🟢 Fully supported.
+- 🟠 Partially supported.
+- 🔴 Not currently supported.
+
+The last column indicates any currently existing GitHub issue for tracking support for this feature.
+Feel free to [create a new issue](https://github.com/coder/envbuilder/issues/new) if you'd like Envbuilder to support a particular feature.
+
+| Status                        | Name                                  | Description                                                                                                            | Known Issues                                           |
```
mtojek, line 14: nit: can we sort them in lex. order?

docs/devcontainer-spec-support.md
johnstcn: I'm finding it difficult to define exactly how we diverge from the spec here. As far as I can tell, we just append `USER $_REMOTE_USER` after the feature directives but we don't appear to use `remoteUser` for anything else.

docs/devcontainer-spec-support.md
```diff
@@ -0,0 +1,72 @@
...
+| 🔴                            | `otherPortsAttributes`                | Other options for ports not configured using `portsAttributes`.                                                        | -                                                      |
+| 🟢                            | `containerEnv`                        | Environment variables to set inside the container.                                                                     | -                                                      |
+| 🟢                            | `remoteEnv`                           | Override environment variables for specific supporting services or tools, but not the container as a whole.            | -                                                      |
+| 🟢                            | `remoteUser`                          | Override the user that supporting tools run as inside the container without changing the user for the whole container. | -                                                      |
```
mtojek, line 23: nit: maybe shorten the description, override user or override env?

docs/devcontainer-spec-support.md
```diff
@@ -0,0 +1,72 @@
...
+- 🟢 Fully supported.
+- 🟠 Partially supported.
+- 🔴 Not currently supported.
+
+The last column indicates any currently existing GitHub issue for tracking support for this feature.
+Feel free to [create a new issue](https://github.com/coder/envbuilder/issues/new) if you'd like Envbuilder to support a particular feature.
+
+| Status                        | Name                                  | Description                                                                                                            | Known Issues                                           |
+| ----------------------------- | ------------------------------------- | ---------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------ |
+| **General**                   |                                       |                                                                                                                        |
```
mtojek, line 16: should we present supported features first?

bpmct (approved):
This is awesome, thanks for putting this together!

Thoughts on also linking to it here? https://containers.dev/supporting

mtojek (approved):

README.md
```diff
@@ -82,19 +82,11 @@ docker run -it --rm
 
 You can see all the supported environment variables in [this document](./docs/env-variables.md).
 
-## Unsupported Features
-
 ### Development Containers
 
-The table below keeps track of features we plan to implement. Feel free to [create a new issue](https://github.com/coder/envbuilder/issues/new) if you'd like Envbuilder to support a particular feature.
+[This document](./docs/supported-features.md) keeps track of what parts of the Dev Container specification Envbuilder currently supports.
```
bpmct, line 87: We need to change it to dev-container-spec-support.md

docs/devcontainer-spec-support.md
```diff
@@ -0,0 +1,72 @@
...
+| 🟢                            | `onCreateCommand`                     | Command to run inside container after first start.                                                                     |                                                        |
+| 🟢                            | `updateContentCommand`                | Command to run after `onCreateCommand` inside container.                                                               |                                                        |
+| 🟢                            | `postCreateCommand`                   | Command to run after `updateContentCommand` inside container.                                                          |                                                        |
+| 🟢                            | `postStartCommand`                    | Command to run each time the container is started.                                                                     |                                                        |
```
bpmct, line 57: We may want to link to an issue showing it is not logged in Coder? Feels a bit strange since this is an individual project

docs/devcontainer-spec-support.md
bpmct: Maybe we should move this doc into `coder/coder` then. Auto-installing VS Code Extensions via `devcontainer.json` will be a top feature request for Coder and I think its important for us to communicate if this use case is possible or not.

Maybe we should say "partially supported" and link to the coder/coder issue and workaround?

johnstcn: @bpmct which specific `coder/coder` issue should we link to? The workaround for installing features is documented in envbuilder#43.

johnstcn: > Maybe we should move this doc into coder/coder then. 

I really don't think this document belongs in `coder/coder`. It's specific to `coder/envbuilder`.

docs/devcontainer-spec-support.md
johnstcn: I intentionally kept the same ordering as https://containers.dev/implementors/json_reference/

docs/devcontainer-spec-support.md
johnstcn: I'm going to punt on ordering for now as it's fairly easy to change in a separate PR.

README.md
johnstcn: whoops!

docs/devcontainer-spec-support.md
johnstcn: I'll add an asterisk (*) for now.

docs/devcontainer-spec-support.md
bpmct: Gotcha. Then let's link to the envbuilder issue/workaround.

mafredri: I don't think coder/coder is the best place either. Just to clarify my POV:

My preferred way to support VS Code customizations would be to "trick" VS Code into believing it is inside a Dev Container. This would make VS Code automatically apply the customizations from the `devcontainer.json`.

Now, should the "tricking" part be a part of envbuilder? Maybe. Or maybe it should be part of the Coder VS Code extension. Or maybe the dev container spec needs to be expanded (via proposal) so that there's an official way to define it. The tool (VS Code) already supports installing extensions, applying settings, the issue is that we're not triggering that mode inside an envbuilder container.

I would like for us to think about the issue of customizations in this manner, **can we get `foobar` to do it by itself?**. Anything outside that scope is IMO the responsibility of `foobar`, a coder module, editor extension, etc.

docs/devcontainer-spec-support.md
mtojek: Ok, let's leave it them. I admit this might be hard to find an interesting item without ctrl+f.

johnstcn: It's a big spec!

Issue #407, opened 2024-11-06
Git SSH Urls cannot select a branch


cmellard:
When using envbuilder with a git ssh url we cannot target a specific branch. Tracing this through it's because of the giturls library succeeds in parsing a transport regex and gleans the url fragment and this is mapped to a branch in coder. However for ssh no fragment is parsed in the scpProvider regex. Presumably it would be good to have a way similar to https to signal to coder a branch we want to check out. If this can't be done via the Git url then perhaps another env var / parameter needs to be available to achieve this

PR #409, opened 2024-11-06
allow an explicit git ref to be passed so we can checkout any ref regardless of the git url 

cmellard:


cmellard:
closes #407

cmellard:
Think this needs more testing. On stopping/starting the workspace looks like the ref / flag isn't persisted

Issue #418, opened 2024-11-18
feat: add improved logging around devcontainer.json lifecycle commands


johnstcn:
# Problem

If you are trying to troubleshoot post create commands etc. it is difficult right now to know if 
a) your postCreateCommand is not getting run, or
b) it is getting run but not emitting any output

# Proposed Solution

Add debug-level logging when building a devcontainer around the following:
- Lifecycle script is not present
- Lifecycle script is present but empty

PR #420, opened 2024-11-22
chore: improve logging around lifecycle scripts

johnstcn:
Closes https://github.com/coder/envbuilder/issues/418

- Adds improved logging around lifecycle scripts and whether we're building from a devcontainer.json or a Dockerfile
- Improves integration tests around lifecycle script execution

### DIFF
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -169,6 +169,7 @@ func run(ctx context.Context, opts options.Options, execArgs *execArgsInfo) erro
 		RemoteEnv:    make(map[string]string),
 	}
 	if fileExists(opts.Filesystem, workingDir.Image()) {
+		opts.Logger(log.LevelInfo, "Found magic image file at %s", workingDir.Image())
 		if err = parseMagicImageFile(opts.Filesystem, workingDir.Image(), &runtimeData); err != nil {
 			return fmt.Errorf("parse magic image file: %w", err)
 		}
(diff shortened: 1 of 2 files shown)


envbuilder.go
```diff
@@ -169,6 +169,7 @@ func run(ctx context.Context, opts options.Options, execArgs *execArgsInfo) erro
 		RemoteEnv:    make(map[string]string),
 	}
 	if fileExists(opts.Filesystem, workingDir.Image()) {
+		opts.Logger(log.LevelInfo, "Found magic image file at %s", workingDir.Image())
```
mafredri, line 172: I'd have gone with debug, but I think your choice of info is better. This'll be helpful in understanding what type of image a user is running.

envbuilder.go
```diff
@@ -334,6 +337,7 @@ func run(ctx context.Context, opts options.Options, execArgs *execArgsInfo) erro
 		} else {
 			// If a Dockerfile was specified, we use that.
 			dockerfilePath := filepath.Join(buildTimeWorkspaceFolder, opts.DockerfilePath)
+			opts.Logger(log.LevelInfo, "Using Dockerfile: %s", opts.DockerfilePath)
```
mafredri, line 340: How about something like "Building in Dockerfile-only mode"?

envbuilder.go
```diff
@@ -1406,6 +1410,7 @@ func execOneLifecycleScript(
 	userInfo userInfo,
 ) error {
 	if s.IsEmpty() {
+		logf(log.LevelInfo, "=== no %s specified", scriptName)
```
mafredri, line 1413:
```suggestion
		logf(log.LevelInfo, "=== No %s specified", scriptName)
```

mafredri (approved):
V nice!

envbuilder.go
johnstcn: Yeah it's way too easy to miss this otherwise!

envbuilder.go
johnstcn: I'll update the corresponding line for "Devcontainer mode" also 👍

PR #434, opened 2024-12-10
fix: use Lchown to prevent erroring on broken symlinks

strike:
It's better to use [Lchown](https://pkg.go.dev/os#Lchown) instead [Chown](https://pkg.go.dev/os#Chown) because:
1. there are dummy/broken symlinks may use in repository. For example, we use dummy/broken symlinks for tests.
2. some software create dummy symlinks in users home. For example, nix with options `--no-channel-add` create broken symlinks:
```
root ➜ / $ ls -l /home/vscode/.nix-defexpr/channels_root
lrwxrwxrwx 1 root root 44 Dec  3 15:34 /home/vscode/.nix-defexpr/channels_root -> /nix/var/nix/profiles/per-user/root/channels
root ➜ / $  ls -l  /nix/var/nix/profiles/per-user/root/channels
ls: cannot access '/nix/var/nix/profiles/per-user/root/channels': No such file or directory
```
Without this PR envbuilder fails with error:
```
chown "/home/vscode": chown /home/vscode/.nix-defexpr/channels_root: no such file or directory
```

### DIFF
--- a/envbuilder.go
+++ b/envbuilder.go
@@ -815,7 +815,7 @@ func run(ctx context.Context, opts options.Options, execArgs *execArgsInfo) erro
 			if err != nil {
 				return err
 			}
-			return os.Chown(path, execArgs.UserInfo.uid, execArgs.UserInfo.gid)
+			return os.Lchown(path, execArgs.UserInfo.uid, execArgs.UserInfo.gid)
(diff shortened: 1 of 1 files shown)


mafredri (approved):
Thanks for fixing this!

PR #436, opened 2024-12-16
feat: allow specifying Kaniko targets for multi-stage dockerfiles

hamishc:
I've encountered issues with large multi-stage dockerfiles when targeting one particular stage. For example, if we have a devcontainer.json like
```
{
    "name": "envbuilder",
    "build": {
		"dockerfile": "Dockerfile",
		"args": {
		},
        "target": "devenv"
	},
}
```
and a Dockerfile as the following
```
FROM alpine AS runtime-deps

FROM runtime-deps AS devenv

FROM runtime-deps AS long-build-stage
COPY some-build-artifact/ /
```
then running `docker run -it -v /my/repo:/workspaces/empty coder/envbuilder:latest` will result in the following error 
```
Failed to build: do build: error building stage: failed to get files used from context: failed to get fileinfo for /workspaces/empty/.devcontainer/some-build-artifact: lstat /workspaces/empty/.devcontainer/some-build-artifact: no such file or directory
Falling back to the default image...
error: do build: error building stage: failed to get files used from context: failed to get fileinfo for /workspaces/empty/.devcontainer/some-build-artifact: lstat /workspaces/empty/.devcontainer/some-build-artifact: no such file or directory: no fallback image has been specified
error: running command "envbuilder": do build: error building stage: failed to get files used from context: failed to get fileinfo for /workspaces/empty/.devcontainer/some-build-artifact: lstat /workspaces/empty/.devcontainer/some-build-artifact: no such file or directory:
    github.com/coder/envbuilder.run.func4
        /home/hamish/code/envbuilder/envbuilder.go:271
  - no fallback image has been specified
```

This is because we don't specify the build target in `KanikoOptions`. Specifying this option can also substantially decrease build times, especially when combined with the `SkipUnusedStages` option - which can allow for large unused build stages to be avoided entirely.


hamishc:
Any chance getting this reviewed sometime soon? This is blocking the adoption of coder for our org, as we rely on multi-stage dockerfiles and devcontainers for most of our repositories.

johnstcn (commented):
Thanks for adding this! Could you add some tests for when this is set?

devcontainer/devcontainer_test.go
```diff
@@ -190,6 +190,27 @@ func TestCompileDevContainer(t *testing.T) {
...
+				Dockerfile: "Dockerfile",
+				Target:     "my-target",
+			},
+		}
+		dcDir := "/workspaces/coder/.devcontainer"
+		err := fs.MkdirAll(dcDir, 0o755)
+		require.NoError(t, err)
+		file, err := fs.OpenFile(filepath.Join(dcDir, "Dockerfile"), os.O_CREATE|os.O_WRONLY, 0o644)
+		require.NoError(t, err)
+		_, err = io.WriteString(file, "FROM scratch AS my-target\nUSER testuser")
```
johnstcn, line 207: Do we want to also add a second target and then assert that we only go up as far as `my-target`? 

e.g.

```
FROM scratch AS my-target
USER testuser
FROM my-target AS not-my-target
USER nottestuser
```

and then assert `params.User` or similar? 

Similarly, what is the expected behaviour for `--skip-unused-stages`?

options/options.go
```diff
@@ -359,6 +363,14 @@ func (o *Options) CLI() serpent.OptionSet {
...
 				"container is being started for the first time.",
 		},
+		{
+			Flag:  "skip-unused-stages",
+			Env:   WithEnvPrefix("SKIP_UNUSED_STAGES"),
+			Value: serpent.BoolOf(&o.SkipUnusedStages),
+			Description: "Skip building all unused docker stages. Otherwise it builds by " +
+				"default all stages, even the unnecessary ones until it reaches the " +
+				"target stage / end of Dockerfile.",
+		},
```
johnstcn, line 373: Where is this actually used?

PR #438, opened 2024-12-18
chore(integration): test with a locally-hosted devcontainer feature

johnstcn:
Modifies integration tests to reference a local devcontainer feature.
This removes the dependency on an external feature and lets us test more interesting things!

### DIFF
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -1545,6 +1545,23 @@ COPY %s .`, testImageAlpine, inclFile)
 func TestPushImage(t *testing.T) {
 	t.Parallel()
 
+	// Write a test feature to an in-memory registry.
+	testFeature := registrytest.WriteContainer(t, registrytest.New(t), "features/test-feature:latest", features.TarLayerMediaType, map[string]any{
+		"install.sh": `#!/bin/sh
+			echo "${MESSAGE}" > /root/message.txt`,
+		"devcontainer-feature.json": features.Spec{
+			ID:      "test-feature",
+			Name:    "test feature",
+			Version: "v0.0.1",
+			Options: map[string]features.Option{
+				"message": {
+					Type:    "string",
+					Default: "hello world",
+				},
+			},
+		},
+	})
+
 	t.Run("CacheWithoutPush", func(t *testing.T) {
 		t.Parallel()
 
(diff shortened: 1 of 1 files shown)


integration/integration_test.go
```diff
@@ -2293,26 +2322,23 @@ RUN date --utc > /root/date.txt`, testImageAlpine),
...
+	t.Run("CacheAndPushDevcontainerFeaturesOverrideOption", func(t *testing.T) {
 		t.Parallel()
 
 		ctx, cancel := context.WithCancel(context.Background())
 		t.Cleanup(cancel)
 
 		srv := gittest.CreateGitServer(t, gittest.Options{
 			Files: map[string]string{
-				// NOTE(mafredri): We can't cache the feature in our local
-				// registry because the image media type is incompatible.
```
mafredri, line 2305: Thanks for figuring this out! ❤️

integration/integration_test.go
```diff
@@ -1603,7 +1623,7 @@ RUN date --utc > /root/date.txt`, testImageAlpine),
 			envbuilderEnv("CACHE_REPO", testRepo),
 			envbuilderEnv("GET_CACHED_IMAGE", "1"),
 		}})
-		require.ErrorContains(t, err, "uncached COPY command is not supported in cache probe mode")
+		require.Regexp(t, `uncached.*?command.*?is not supported in cache probe mode`, err.Error())
```
mafredri, line 1626:
```suggestion
		require.Regexp(t, `uncached.*command.*is not supported in cache probe mode`, err.Error())
```
The `?` is superfluous as `*` already allows zero or more.

integration/integration_test.go
```diff
@@ -1702,7 +1731,7 @@ RUN date --utc > /root/date.txt`, testImageAlpine),
 		_, err = runEnvbuilder(t, runOpts{env: append(opts,
 			envbuilderEnv("GET_CACHED_IMAGE", "1"),
 		)})
-		require.ErrorContains(t, err, "error probing build cache: uncached COPY command")
+		require.Regexp(t, "error probing build cache: uncached.*?command.*?is not supported in cache probe mode", err.Error())
```
mafredri, line 1734:
```suggestion
		require.Regexp(t, "error probing build cache: uncached.*command.*is not supported in cache probe mode", err.Error())
```

mafredri (approved):
Looks great, nice work!

integration/integration_test.go
johnstcn: It kind of is, but it also makes the match non-greedy. This is just a reflexive thing for me as I find it makes the expression less likely to accidentally consume all of the input.

PR #439, opened 2024-12-19
refactor(testutil): refactor registrytest

johnstcn:
Refactors registrytest to a form that can be used by both envbuilder and terraform-provider-envbuilder.
Also fixes an issue with `url.Parse` incorrectly interpreting the scheme of `host:port` and attempting to push an image to Docker hub.


mafredri (approved):
Changes look good 👍🏻. Re: PR title either `test` or `refactor` would suit better as `chore` is for non-code related changes (although it's sometimes used "incorrectly" when something could be considered a chore).

PR #446, opened 2025-01-11
feat: add git clone with thinpack option to support self hosted azure devops

CH-Chang:
The program uses only the hostname to address the issue of requiring thin pack to perform a git clone operation on Azure DevOps, as shown in the code snippet below.

https://github.com/coder/envbuilder/blob/d045c1ce27a5995ac60409bd48ad8e90861b6bfd/git/git.go#L56-L78

However, since Azure DevOps supports self-hosted servers, the hostname may not necessarily be dev.azure.com. Therefore, it is desired to add a command-line option to enable this functionality.

### DIFF
--- a/docs/env-variables.md
+++ b/docs/env-variables.md
@@ -27,7 +27,7 @@
 | `--git-url` | `ENVBUILDER_GIT_URL` |  | The URL of a Git repository containing a Devcontainer or Docker image to clone. This is optional. |
 | `--git-clone-depth` | `ENVBUILDER_GIT_CLONE_DEPTH` |  | The depth to use when cloning the Git repository. |
 | `--git-clone-single-branch` | `ENVBUILDER_GIT_CLONE_SINGLE_BRANCH` |  | Clone only a single branch of the Git repository. |
-| `--git-clone-thinpack` | `ENVBUILDER_GIT_CLONE_THINPACK` |  | Git clone with thin pack compatibility enabled, ensuring that even when thin pack compatibility is activated, it will not be turned on for the domain dev.zaure.com.  |
+| `--git-clone-thinpack` | `ENVBUILDER_GIT_CLONE_THINPACK` | `true` | Git clone with thin pack compatibility enabled, ensuring that even when thin pack compatibility is activated,it will not be turned on for the domain dev.zaure.com. |
 | `--git-username` | `ENVBUILDER_GIT_USERNAME` |  | The username to use for Git authentication. This is optional. |
 | `--git-password` | `ENVBUILDER_GIT_PASSWORD` |  | The password to use for Git authentication. This is optional. |
 | `--git-ssh-private-key-path` | `ENVBUILDER_GIT_SSH_PRIVATE_KEY_PATH` |  | Path to an SSH private key to be used for Git authentication. If this is set, then GIT_SSH_PRIVATE_KEY_BASE64 cannot be set. |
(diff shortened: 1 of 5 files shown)


git/git.go
```diff
@@ -53,7 +54,7 @@ func CloneRepo(ctx context.Context, logf func(string, ...any), opts CloneRepoOpt
 		return false, fmt.Errorf("parse url %q: %w", opts.RepoURL, err)
 	}
 	logf("Parsed Git URL as %q", parsed.Redacted())
-	if parsed.Hostname() == "dev.azure.com" {
+	if parsed.Hostname() == "dev.azure.com" || opts.ThinPack {
```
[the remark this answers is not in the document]
johnstcn: I think the semantics of the option as currently defined are also somewhat misleading. Setting `ENVBUILDER_GIT_CLONE_THINPACK=true` will add thinpack to the _unsupported_ capabilities of the Git transport when attempting to clone. Shouldn't it be the other way round?

docs/env-variables.md
```diff
@@ -27,6 +27,7 @@
 | `--git-url` | `ENVBUILDER_GIT_URL` |  | The URL of a Git repository containing a Devcontainer or Docker image to clone. This is optional. |
 | `--git-clone-depth` | `ENVBUILDER_GIT_CLONE_DEPTH` |  | The depth to use when cloning the Git repository. |
 | `--git-clone-single-branch` | `ENVBUILDER_GIT_CLONE_SINGLE_BRANCH` |  | Clone only a single branch of the Git repository. |
+| `--git-clone-thinpack` | `ENVBUILDER_GIT_CLONE_THINPACK` |  | Clone with thin pack compabilities. |
```
johnstcn, line 30: Can you add a note regarding the behaviour of thinpack here for repos hosted on `dev.zaure.com`?

CH-Chang: I have added the description regarding the behavior of thinpack for repositories hosted on dev.zaure.com, where the option remains ineffective even when enabled. Thank you!

git/git.go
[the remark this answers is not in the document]
CH-Chang: > Thanks for this! Would you mind updating the logf message below? It still states that this is for azure specifically, which could be misleading if ThinPack were set in non Azure environments.

I've updated the log message to differentiate between disabling the feature via the command line and disabling it due to encountering the corresponding domain.

> I think the semantics of the option as currently defined are also somewhat misleading. Setting `ENVBUILDER_GIT_CLONE_THINPACK=true` will add thinpack to the _unsupported_ capabilities of the Git transport when attempting to clone. Shouldn't it be the other way round?

I've adjusted the definition of the option. By default, the option is set to true, which enables Git clone thin pack compatibility.

git/git.go
```diff
@@ -53,28 +54,40 @@ func CloneRepo(ctx context.Context, logf func(string, ...any), opts CloneRepoOpt
...
+			//
+			// New commits and pushes against a remote worked without any issues.
+			// See: https://github.com/go-git/go-git/issues/64
+			//
+			// This is knowingly not safe to call in parallel, but it seemed
+			// like the least-janky place to add a super janky hack.
+			thinPack = false
+			logf("Workaround for Azure DevOps: marking thin-pack as unsupported")
+		}
+	}
```
johnstcn, line 85: nit: `else if { ... }`

johnstcn (approved):
Thanks!, Just a couple of nits on my end, but nothing blocking. 

You may need to re-run `make docs/env-variables.md` to appease CI.

git/git.go
CH-Chang: Thanks, I have updated the code to use else if { ... } and have run make docs/env-variables.md to update the documentation.

SasSwart (approved):

johnstcn:
Thanks @CH-Chang !

johnstcn:
Created https://github.com/coder/terraform-provider-envbuilder/issues/89 to support this in the provider.

PR #485, opened 2025-12-11
feat: add git submodule support

### LINKED ISSUE #74: Git submodule support
There doesn't appear to be a way to clone a repo with submodules. In my use case, the devcontainer is built up using files coming in from a submod, so this is critical.

It looks looks like support for this would need to be added to [`CloneRepoOptions`](https://github.com/coder/envbuilder/blob/8aec7c97ba6838801982da3c361c7d83e56f49a9/git.go#L18) in `git.go`. The underlying `git-go` options already support this, so hopefully this isn't a heavy lift. See [`CloneOptions.RecurseSubmodules`](https://github.com/go-git/go-git/blob/eab7337c1d22e893bdd0ad731b2d6d4676b289b9/options.go#L64) in `go-git/options.go`.

Finding the above references is about the extent of my Go knowledge though, so I can't offer a patch myself.

bjornrobertsson:
### Summary

This PR resolves/mitigates https://github.com/coder/envbuilder/issues/74

### Changes

Since the existing submodule support in `git.go` depends on execution of the git binary, this implementation circumvents that limitation by traversing the tree to clone submodules natively in Go.

### Behavior

- Functionality can be toggled via a parameter (ENVBUILDER_GIT_CLONE_SUBMODULES)
- Submodule initialization runs on every workspace start when enabled, so a transient failure during the first run is automatically retried on the next start
- Already-cloned submodules are detected and only their expected commit is checked out, making re-runs cheap


johnstcn (commented):
Thanks for working on this @bjornrobertsson ! 
I have some comments below, and I think it would also be worthwhile adding an integration test (see: `integration/integration_test.go` and `testutil/gittest/gittest.go`).

johnstcn (approved):

git/git.go
```diff
@@ -419,3 +433,276 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
 		done:        done,
 	}
 }
+
+// resolveSubmoduleURL resolves a potentially relative submodule URL against the parent repository URL
```
mafredri:
```suggestion
```

git/git.go
```diff
@@ -419,3 +433,276 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
...
+	cfg, err := repo.Config()
+	if err != nil {
+		return fmt.Errorf("get repo config: %w", err)
+	}
+
+	parentURL := opts.RepoURL
+	if origin, hasOrigin := cfg.Remotes["origin"]; hasOrigin && len(origin.URLs) > 0 {
+		parentURL = origin.URLs[0]
+	}
+	logf("Parent repository URL: %s", parentURL)
```
mafredri: Should we consider logging the redacted URL here? This URL could contain credentials.

git/git.go
```diff
@@ -419,3 +433,276 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
...
+// resolveSubmoduleURL resolves a potentially relative submodule URL against the parent repository URL
+// ResolveSubmoduleURL resolves a potentially relative submodule URL against a parent repository URL.
+func ResolveSubmoduleURL(parentURL, submoduleURL string) (string, error) {
+	// If the submodule URL is absolute (contains ://) or doesn't start with ./ or ../, return it as-is
+	if strings.Contains(submoduleURL, "://") || (!strings.HasPrefix(submoduleURL, "../") && !strings.HasPrefix(submoduleURL, "./")) {
+		return submoduleURL, nil
+	}
+
+	// Parse the parent URL
+	parentParsed, err := url.Parse(parentURL)
```
mafredri: If the parent URL is SCP-like (`git@github.com:org/repo.git`), and the submodule is relative, we don't hit the guard above and `url.Parse` can't parse the SCP-like URL.

Suggestion: Use `parentEP := transport.NewEndpoint(parentURL)` instead? You should be able to just update the `parentEP.Path` and return a string instead of the `url.URL` construction below.

testutil/gittest/gittest.go
```diff
@@ -270,6 +270,40 @@ func NewRepo(t *testing.T, fs billy.Filesystem, commits ...CommitFunc) *git.Repo
...
+	for path, content := range opts.Files {
+		commits = append(commits, Commit(t, path, content, "my test commit"))
+	}
+	// Add gitmodules file pointing to the submodule server
+	gitmodulesContent := fmt.Sprintf(`[submodule "submod"]
+	path = submod
+	url = %s
+`, submoduleSrv.URL)
+	commits = append(commits, Commit(t, ".gitmodules", gitmodulesContent, "add submodule"))
+	_ = NewRepo(t, fs, commits...)
```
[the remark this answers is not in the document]
bjornrobertsson: @mafredri can we do something to align that? I know the code 'works', the customer has been using these patches or the premises made here for some time. But I am aware I don't approach it from Eng side, more 'what works'.

mafredri (commented):
First-pass review (Netero). These are mechanical findings only; the full review panel has not yet reviewed this PR. The panel will review after these findings are addressed.

Severity count: 1 P1 (security), 1 P2 (logic bug), 1 P3 (consistency), 1 Nit.

The submodule feature addresses a real gap and the overall structure is reasonable. The security guard (`SameHost`) and URL redaction are good additions. However, the credential-forwarding logic has a bypass path through nested recursion, and the relative-URL resolution has a semantic error for `./`-prefixed paths.

> "Attack path: a malicious .gitmodules on github.com points submodule A to evil.com. The first-level SameHost check blocks auth. But if evil.com/repo itself has a nested submodule also on evil.com, the second-level check passes and forwards the root GitHub PAT to evil.com." *(Netero)*

git/git.go
```diff
@@ -419,3 +434,350 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
...
+		}
+
+		// Check for nested submodules
+		subWorktree, err := subRepo.Worktree()
+		if err == nil {
+			nestedSubs, err := subWorktree.Submodules()
+			if err == nil && len(nestedSubs) > 0 {
+				logf("  Found %d nested submodule(s) in %s", len(nestedSubs), subConfig.Name)
+				// Create new opts with the submodule's URL as the parent
+				nestedOpts := opts
```
mafredri: **P1** [DEREM-1] Root repository credentials leak to nested submodules on untrusted hosts.

When auth is withheld at depth 1 (different host), the recursive call at this line copies `opts` which still carries `opts.RepoAuth` (the root's credentials). At depth 2, `SameHost(nestedOpts.RepoURL, nestedResolvedURL)` compares the evil host against itself, passes, and forwards the root PAT.

Attack scenario: `.gitmodules` on `github.com` points submodule A to `evil.com/repo`. Auth correctly blocked. But `evil.com/repo` has a nested submodule also on `evil.com`. The depth-2 `SameHost` check sees matching hosts and leaks the root credential.

Fix: when `cloneSubmodule` does NOT forward auth, set `nestedOpts.RepoAuth = nil` before recursing. Or track the root URL explicitly and always compare against it. *(Netero)*

> 🤖

git/git.go
```diff
@@ -419,3 +434,350 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
...
+	// Split the submodule URL into components
+	// and manually walk up the directory tree for each ../
+	currentPath := parentPath
+	relativeParts := strings.Split(submoduleURL, "/")
+
+	for _, part := range relativeParts {
+		if part == ".." {
+			// Go up one directory
+			currentPath = path.Dir(currentPath)
+		} else if part == "." {
```
mafredri, line 561: **P2** [DEREM-2] `ResolveSubmoduleURL` mishandles `./`-relative URLs.

The `.` case does `continue` (no-op), leaving `currentPath` at the full parent path including the repo "filename." For `./extras/tool.git` relative to `https://example.com/org/main.git`, this produces `https://example.com/org/main.git/extras/tool.git`.

Git's behavior: both `.` and `..` are relative to the *directory* containing the repo, not the repo path itself. The correct result is `https://example.com/org/extras/tool.git`. The `..` case works by accident because `path.Dir` strips the filename AND goes up one level in a single operation.

Fix: before the loop, set `currentPath = path.Dir(parentPath)` to establish directory context. Then `..` means "go up from directory" and `.` means "stay."

The test `TestResolveSubmoduleURL/relativeChild` asserts the wrong expected value, so it passes vacuously. *(Netero)*

> 🤖

git/git.go
```diff
@@ -419,3 +434,350 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
...
+			RefSpecs: []config.RefSpec{
+				config.RefSpec("+" + expectedHash.String() + ":" + expectedHash.String()),
+			},
+			Auth:            submoduleAuth,
+			Progress:        opts.Progress,
+			InsecureSkipTLS: opts.Insecure,
+			CABundle:        opts.CABundle,
+			ProxyOptions:    opts.ProxyOptions,
+		})
+		if err != nil && err != git.NoErrAlreadyUpToDate {
```
mafredri: **P3** [DEREM-3] Bare `!=` comparison against sentinel error instead of `errors.Is`.

Lines 745 and 756 use `err != git.NoErrAlreadyUpToDate`. The same file uses `errors.Is` at lines 113, 134, and 724 for other go-git sentinels. If `NoErrAlreadyUpToDate` is ever wrapped, this comparison silently fails.

Fix: `!errors.Is(err, git.NoErrAlreadyUpToDate)` at both locations. *(Netero)*

> 🤖

git/git_test.go
```diff
@@ -533,6 +533,349 @@ func mustRead(t *testing.T, fs billy.Filesystem, path string) string {
...
+		},
+		{
+			name:   "file URL",
+			input:  "file:///local/repo.git",
+			expect: "file:///local/repo.git",
+		},
+	}
+
+	for _, tc := range cases {
+		tc := tc
```
mafredri: **Nit** [DEREM-4] Unnecessary `tc := tc` loop variable capture.

Three new instances (lines 674, 772, 848). Go 1.24 makes per-iteration loop variable scoping the default. These are dead code. *(Netero)*

> 🤖

git/git.go
```diff
@@ -419,3 +434,350 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
...
+	} else if opts.RepoAuth != nil {
+		logf("  ⚠ Not forwarding auth to submodule (different host: %s)", extractHost(resolvedURL))
+	}
+
+	// Check if already cloned
+	_, err = subFS.Stat(".git")
+	if err == nil {
+		logf("  Submodule already cloned, checking out expected commit...")
+		// Open the existing repository
+		subRepo, err := git.Open(
```
mafredri, line 731: **P2** [DEREM-5] "Already cloned" path passes worktree root as git storage, guaranteeing failure.

`filesystem.NewStorage(subFS, ...)` here tells go-git to look for `HEAD`, `objects/`, `refs/` at the submodule worktree root. The actual git metadata lives at `subFS/.git/`. Compare with the fresh-clone path 40 lines below which correctly does `subFS.Chroot(".git")` before creating storage.

Consequence: any crash recovery or partial-prior-run scenario where `.git` already exists enters this path and fails with a misleading "repository does not exist" error, aborting all remaining submodules.

Fix:
```go
subGitDir, err := subFS.Chroot(".git")
if err != nil {
    return fmt.Errorf("chroot to .git: %w", err)
}
subRepo, err := git.Open(
    filesystem.NewStorage(subGitDir, cache.NewObjectLRU(cache.DefaultMaxSize*10)),
    subFS,
)
```

*(Hisoka P2, Kite P2, Ryosuke P2, Meruem P2, Razor P3)*

> 🤖

git/git.go
```diff
@@ -132,6 +137,15 @@ func CloneRepo(ctx context.Context, logf func(string, ...any), opts CloneRepoOpt
 	if err != nil {
 		return false, fmt.Errorf("clone %q: %w", opts.RepoURL, err)
 	}
+
```
mafredri, line 146: **P2** [DEREM-6] Submodule init failure is permanently unrecoverable.

If any submodule clone fails (network timeout, transient auth issue), `CloneRepo` returns `(true, error)`. On the next workspace start, the parent repo already exists, so `CloneRepo` hits line 119 (`repo != nil`) and returns `(false, nil)`. Submodule initialization is never retried.

Consequence: a single network blip during first workspace setup creates a workspace that permanently lacks submodule content. Recovery requires deleting the entire workspace folder.

Consider: check for pending submodules even when the repo already exists, or store a marker file indicating incomplete submodule init that triggers a retry. *(Mafuuu P2, Takumi P2)*

> 🤖

git/git.go
```diff
@@ -419,3 +434,350 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
...
+			continue
+		} else if part != "" {
+			// Add this component to the path
+			currentPath = currentPath + "/" + part
+		}
+	}
+
+	// Reconstruct the URL with the resolved path.
+	parentEP.Path = path.Clean(currentPath)
+	return parentEP.String(), nil
```
mafredri, line 572: **P2** [DEREM-7] Credential leak via `Endpoint.String()` re-emitting parent URL credentials.

`transport.NewEndpoint(parentURL)` parses embedded credentials from the parent URL. `parentEP.String()` at this line re-emits them (confirmed: go-git `transport/common.go:179-183` writes `url.PathEscape(u.User)` and `url.PathEscape(u.Password)` into the output).

The resolved URL then flows to `cloneSubmodule` as the clone target and to `nestedOpts.RepoURL` for recursive calls. The `SameHost` gate controls only the `RepoAuth` path; URL-embedded credentials are an uncontrolled parallel channel. If a parent URL is `https://token@github.com/org/repo` and a submodule resolves to a different host, the token still appears in the resolved URL.

Fix: zero out `parentEP.User` and `parentEP.Password` before calling `String()`. Credentials should flow exclusively through `RepoAuth`. *(Kurapika P2)*

> 🤖

git/git.go
```diff
@@ -419,3 +434,350 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
...
+		// Clone the submodule manually
+		err = cloneSubmodule(ctx, logf, w, subConfig, subStatus.Expected, resolvedURL, opts)
+		if err != nil {
+			return fmt.Errorf("clone submodule %q: %w", subConfig.Name, err)
+		}
+
+		logf("✓ Submodule initialized: %s", subConfig.Name)
+
+		// Recursively handle nested submodules
+		subRepo, err := sub.Repository()
```
mafredri: **P2** [DEREM-8] Nested submodule recursion silently fails because `sub.Repository()` requires initialization the custom clone bypasses.

Go-git's `Submodule.Repository()` checks `s.initialized` (set only when the submodule appears in the parent repo's `.git/config`). In a freshly cloned repo, `.git/config` contains only the origin remote. Submodule entries appear only after `sub.Init()` (or `git submodule init`). The custom `cloneSubmodule` clones content to the filesystem but never calls `sub.Init()`, so `sub.Repository()` returns `ErrSubmoduleNotInitialized`, the `continue` fires, and nested submodules are silently skipped.

Consequence: `SubmoduleDepth > 1` is advertised (default 10 for "true") but depth > 1 is non-functional. The integration test only covers flat submodules.

Fix: either call `sub.Init()` before recursing, or open the nested repo directly from the filesystem path via `git.Open(storage, fs)` on the chrooted directory. *(Pariston P2)*

> 🤖

git/git.go
```diff
@@ -419,3 +434,350 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
...
+	subGitDir, err := subFS.Chroot(".git")
+	if err != nil {
+		return fmt.Errorf("chroot to .git: %w", err)
+	}
+
+	gitStorage := filesystem.NewStorage(subGitDir, cache.NewObjectLRU(cache.DefaultMaxSize*10))
+
+	// Clone the submodule repository
+	// Use SingleBranch=false to fetch all branches so we can find the commit
+	subRepo, err := git.CloneContext(ctx, gitStorage, subFS, &git.CloneOptions{
```
mafredri, line 759: **P2** [DEREM-9] Submodule clone fetches unbounded history of all branches without depth limit.

The parent clone respects `opts.Depth` for shallow cloning. Submodule clones set `SingleBranch: false` and pass no `Depth`. A submodule pointing at a large repository downloads complete history across all branches.

Consequence: workspace startup that takes seconds for a shallow main clone can take minutes or GB of bandwidth per large submodule. Since `.gitmodules` is a committed file, a malicious contributor on a trusted host can add a submodule pointing at any massive public repo.

Fix: propagate `opts.Depth` (or a submodule-specific default like 1) into the submodule `CloneOptions`. The commit-verification fetch can deepen if the expected hash isn't in the shallow clone. *(Hisoka P2)*

> 🤖

git/git_test.go
```diff
@@ -533,6 +533,349 @@ func mustRead(t *testing.T, fs billy.Filesystem, path string) string {
...
+				require.ErrorContains(t, err, c.expectErr)
+				return
+			}
+			require.NoError(t, err)
+			require.Equal(t, c.expect, got)
+		})
+	}
+}
+
+func TestCloneOptionsFromOptions_Submodules(t *testing.T) {
```
mafredri, line 895: **P2** [DEREM-10] Core submodule logic (180 lines) has no unit test coverage beyond one integration happy path.

`cloneSubmodule` has a three-stage retry chain (clone, verify commit, fetch by hash, fetch all, verify again), depth-limiting, credential-forwarding decisions, and an "already cloned" early-return. None of these are tested.

The integration test hits one path: single submodule, no auth, same host, commit available on first clone, no nesting. If the retry logic breaks, depth limit off-by-ones, or credential gating misfires, no test catches it.

Suggested additions: internal test file exercising retry (submodule whose commit requires fetch), depth limiting (nested submodule at max depth), and credential non-forwarding (authenticated parent with different-host submodule). *(Bisky P2, Chopper P3)*

> 🤖

git/git.go
```diff
@@ -419,3 +434,350 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
...
+		URL:             resolvedURL,
+		Auth:            submoduleAuth,
+		Progress:        opts.Progress,
+		InsecureSkipTLS: opts.Insecure,
+		CABundle:        opts.CABundle,
+		ProxyOptions:    opts.ProxyOptions,
+		SingleBranch:    false, // Fetch all branches
+		NoCheckout:      true,  // Don't checkout yet, we'll do it manually
+	})
+	if err != nil && !errors.Is(err, git.ErrRepositoryAlreadyExists) {
```
mafredri: **P3** [DEREM-11] Nil-pointer dereference if `CloneContext` returns `ErrRepositoryAlreadyExists`.

The guard allows `ErrRepositoryAlreadyExists` through, leaving `subRepo` nil. Line 730 then calls `subRepo.CommitObject(expectedHash)`, which panics. Currently unreachable (`.git` is freshly created at line 700), but the explicit defensive check signals the author anticipated this case without completing the handling.

Fix: either remove the `ErrRepositoryAlreadyExists` exception, or open the existing repo when it triggers. *(Mafuuu P3, Chopper P2, Takumi P3, Razor P3)*

> 🤖

git/git.go
```diff
@@ -419,3 +434,350 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
...
+// This handles: git@github.com:org/repo, deploy@host:repo, user@10.0.0.5:project
+var scpLikeURLRegex = regexp.MustCompile(`^([^@]+)@([^:]+):(.+)$`)
+
+// extractHost extracts the host from a URL, handling both standard URLs and SCP-like URLs.
+// Returns empty string if host cannot be determined.
+func extractHost(u string) string {
+	// Try standard URL parsing first
+	if parsed, err := url.Parse(u); err == nil && parsed.Host != "" {
+		// Remove port if present
+		host := parsed.Hostname()
```
mafredri, line 459: **P3** [DEREM-12] `SameHost` strips port, forwarding credentials to any service sharing the hostname.

`extractHost` uses `parsed.Hostname()` which discards port. If a parent is on `corp.com:443` and a submodule points to `corp.com:9999` (different service), credentials are forwarded. This matches standard git credential-helper behavior (host-only matching), so it follows ecosystem convention, but it's a broader trust surface than necessary for a feature that explicitly gates credential forwarding.

Consider: include port in comparison when explicitly specified, or document this as intentional. *(Hisoka P3, Kurapika P3, Knov P3)*

> 🤖

git/git.go
```diff
@@ -419,3 +434,350 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
...
+		return fmt.Errorf("get repo config: %w", err)
+	}
+
+	parentURL := opts.RepoURL
+	if origin, hasOrigin := cfg.Remotes["origin"]; hasOrigin && len(origin.URLs) > 0 {
+		parentURL = origin.URLs[0]
+	}
+	logf("Parent repository URL: %s", RedactURL(parentURL))
+
+	for _, sub := range subs {
```
mafredri, line 608: **P3** [DEREM-13] No context cancellation check between submodule iterations.

The loop iterates without checking `ctx.Done()`. If context is canceled mid-iteration (workspace teardown, timeout), the code enters the next submodule's setup before any network call detects the cancellation. For repos with many submodules, this delays shutdown.

Fix: `select { case <-ctx.Done(): return ctx.Err() default: }` at the top of the loop body. *(Ryosuke P3, Knov P3)*

> 🤖

options/options.go
```diff
@@ -108,6 +148,10 @@ type Options struct {
 	GitCloneSingleBranch bool
 	// GitCloneThinPack clone with thin pack compabilities. This is optional.
 	GitCloneThinPack bool
+	// GitCloneSubmodules controls submodule initialization after cloning.
+	// 0 = disabled (default), positive integer = max recursion depth.
+	// Accepts "true" (defaults to 10), "false" (0), or a positive integer.
```
mafredri: **P3** [DEREM-14] `GitCloneSubmodules int` reads as a boolean toggle, not a depth.

A reader encountering `opts.GitCloneSubmodules = 5` guesses "5 submodules" not "5 levels of recursion." The internal API uses `SubmoduleDepth` which is clear. Consider `GitCloneSubmoduleDepth` or `GitSubmoduleDepth` for the options field to align with the internal type. *(Gon P3)*

> 🤖

git/git.go
```diff
@@ -419,3 +434,350 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
...
+		if err != nil {
+			return fmt.Errorf("get submodule worktree: %w", err)
+		}
+
+		// Checkout the expected commit
+		err = subWorktree.Checkout(&git.CheckoutOptions{
+			Hash: expectedHash,
+		})
+		if err != nil {
+			return fmt.Errorf("checkout expected commit: %w", err)
```
mafredri: **P3** [DEREM-15] Error message drops the commit hash on the "already cloned" checkout path.

Line 779 (fresh clone path) includes the hash: `"checkout expected commit %s: %w"`. Line 691 (same operation, existing repo) omits it: `"checkout expected commit: %w"`. The operator seeing this error has to hunt for which commit was expected.

Fix: `fmt.Errorf("checkout expected commit %s: %w", expectedHash, err)` *(Leorio P3)*

> 🤖

git/git.go
```diff
@@ -419,3 +434,350 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
...
+		// Recursively handle nested submodules
+		subRepo, err := sub.Repository()
+		if err != nil {
+			logf("  ⚠ Could not open submodule repository %s: %v", subConfig.Name, err)
+			continue
+		}
+
+		// Check for nested submodules
+		subWorktree, err := subRepo.Worktree()
+		if err == nil {
```
mafredri: **P3** [DEREM-16] Silent error when getting nested submodule worktree.

If `subRepo.Worktree()` at line 628 returns an error, execution falls through silently with no log. Compare with the similar error at line 621-623 which logs a warning. A worktree error means corrupted submodule state silently skips nested initialization with no diagnostic.

Fix: add `logf("  ⚠ Could not get worktree for %s: %v", subConfig.Name, err)` in an else branch. *(Razor P3)*

> 🤖

git/git_test.go
```diff
@@ -533,6 +533,349 @@ func mustRead(t *testing.T, fs billy.Filesystem, path string) string {
...
+
+func TestResolveSubmoduleURL(t *testing.T) {
+	t.Parallel()
+
+	cases := []struct {
+		name      string
+		parentURL string
+		subURL    string
+		expect    string
+		expectErr string
```
mafredri, line 787: **P3** [DEREM-17] `expectErr` field declared but unused; error path of `ResolveSubmoduleURL` untested.

The struct declares `expectErr string` and the loop checks it, but every row leaves it empty. The error path (`transport.NewEndpoint` failure) has no coverage. Add a case with an unparseable parent URL (e.g., `"://\x00"`) to exercise the error branch. *(Bisky P3, Chopper Nit)*

> 🤖

options/options.go
```diff
@@ -5,13 +5,53 @@ import (
...
 )
 
+// SubmoduleDepth is a custom type for handling submodule depth that accepts
+// "true" (defaults to 10), "false" (0), or a positive integer.
+type SubmoduleDepth int
+
+const DefaultSubmoduleDepth = 10
+
+func (s *SubmoduleDepth) Set(val string) error {
+	lower := strings.ToLower(strings.TrimSpace(val))
```
mafredri, line 23: **Nit** [DEREM-18] `strconv.Atoi` uses the untrimmed original `val`, not `lower`.

`TrimSpace` benefits the switch cases but if the env value is `" 5 "` (copy-paste artifact), Atoi fails even though the intent is clear. Use `lower` or trim `val` before Atoi. *(Hisoka)*

> 🤖

options/options_test.go
```diff
@@ -72,6 +72,37 @@ func TestEnvOptionParsing(t *testing.T) {
 			require.False(t, o.GitCloneSingleBranch)
 			require.True(t, o.GitCloneThinPack)
 		})
+
+		t.Run("remote repo build mode", func(t *testing.T) {
```
mafredri, line 76: **Nit** [DEREM-19] Submodule depth tests are nested under the "bool" subtest parent.

`SubmoduleDepth` is a custom type accepting integers, "true", and "false". Nesting under `bool/` makes test output misleading (`TestEnvOptionParsing/bool/git_clone_submodules_depth`). Move to a sibling `t.Run` at the same level as "string", "int", "bool". *(Bisky, Chopper)*

> 🤖

mafredri (dismissed):
Full panel review (16 reviewers). The submodule feature fills a real gap and the security-conscious design (SameHost gating, URL redaction) shows good instincts. However, the implementation has several structural issues that need attention before merge.

Severity count: 1 P1, 7 P2, 8 P3, 3 Nit (including R1 findings).

The most critical cluster: the "already cloned" recovery path uses the wrong filesystem root for git storage (5 reviewers independently found this). Combined with the fact that submodule init never retries on subsequent starts, any transient failure during first workspace setup permanently breaks submodule state with no automatic recovery.

A second cluster around credential handling: beyond the R1 credential-leak-through-recursion (DEREM-1), the panel found that `Endpoint.String()` re-emits URL-embedded credentials into resolved URLs, creating a parallel credential leak channel that bypasses the `SameHost` gate entirely.

The panel also identified that nested submodule recursion (depth > 1) is likely non-functional because `sub.Repository()` requires the submodule to be registered in `.git/config` (via `sub.Init()`), which the custom clone path never does. The depth parameter accepts values up to MaxInt but depth > 1 silently produces warning logs and skips.

Architectural note: multiple reviewers observed that a thinner wrapper around go-git's existing `Submodule.UpdateContext()` (addressing only the two narrow gaps: relative URL resolution and per-submodule auth) would avoid the class of bugs found here. The 200-line reimplementation reintroduces problems that go-git's built-in API already solves (storage layout, initialization tracking, recursive update). This is not blocking, but worth considering.

> "An attacker controlling .gitmodules on github.com points submodule A to evil.com. Auth correctly blocked at depth 1. But evil.com/repo has a nested submodule also on evil.com. The depth-2 SameHost check sees matching hosts and leaks the root credential." *(Netero)*

> "If go-git's internal expectations differ from the manual clone layout, nested recursion silently fails at the continue on line 623 and nested submodules are never initialized." *(Hisoka)*

git/git.go
```diff
@@ -418,3 +432,271 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
 		done:        done,
 	}
 }
+
+// resolveSubmoduleURL resolves a potentially relative submodule URL against the parent repository URL
+// ResolveSubmoduleURLForTest is exported for testing resolveSubmoduleURL logic
+func ResolveSubmoduleURLForTest(parentURL, submoduleURL string) (string, error) {
```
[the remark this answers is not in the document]
mafredri: The function was renamed to `ResolveSubmoduleURL` in commit db5e40d on this PR.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git_test.go
```diff
@@ -492,6 +492,74 @@ func mustRead(t *testing.T, fs billy.Filesystem, path string) string {
...
+			parentURL: "https://example.com/org/main.git",
+			subURL:    "./extras/tool.git",
+			expect:    "https://example.com/org/main.git/extras/tool.git",
+		},
+		{
+			name:      "badParent",
+			parentURL: "://bad",
+			subURL:    "./child",
+			expectErr: "parse parent URL",
+		},
```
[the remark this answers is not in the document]
mafredri: `TestResolveSubmoduleURL` now covers `ssh://` parents (with both user-preserved and password-stripped cases), `git://` parents, and an unparseable URL that exercises the error path. SCP-like (`git@host:path`) was already covered.

The remaining schemes from `TestSetupRepoAuth` (absolute and bare relative local paths) short-circuit at the top of `ResolveSubmoduleURL` because they do not start with `./` or `../`, so they have no resolver-specific behavior to test.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
```diff
@@ -136,6 +140,15 @@ func CloneRepo(ctx context.Context, logf func(string, ...any), opts CloneRepoOpt
...
 		return false, fmt.Errorf("clone %q: %w", opts.RepoURL, err)
 	}
+
+	// Initialize submodules if requested
+	if opts.Submodules {
+		err = initSubmodules(ctx, logf, repo, opts)
+		if err != nil {
+			return true, fmt.Errorf("init submodules: %w", err)
+		}
+	}
```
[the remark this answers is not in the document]
mafredri: Settled in-thread: @johnstcn agreed to keep the explicit opt-in so behavior aligns with `git clone` rather than auto-detecting via `.gitmodules`. No further action.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

README.md
```diff
@@ -118,3 +118,10 @@ On macOS or Windows systems, we recommend using a VM or the provided `.devcontai
 - `test`: Runs tests.
 - `test-registry`: Stands up a local registry for caching images used in tests.
 - `docs/env-variables.md`: Updated the [environment variables documentation](./docs/env-variables.md).
+  
+**Submodule Handling Fix**
+
+An issue concerning git's submodule handling has been resolved through iterative refinements. This fix ensures robust submodule cloning and URL resolution without relying on the calls to the git binary (current fallback).
+```
+ENVBUILDER_GIT_CLONE_SUBMODULES=true
```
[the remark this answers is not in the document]
mafredri: Removed from `README.md` in commit 442477e on this PR.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
```diff
@@ -418,3 +432,271 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
...
+		User:   parentParsed.User,
+		Host:   parentParsed.Host,
+		Path:   resolvedPath,
+	}
+
+	return resolvedParsed.String(), nil
+}
+
+// initSubmodules recursively initializes and updates all submodules in the repository.
+func initSubmodules(ctx context.Context, logf func(string, ...any), repo *git.Repository, opts CloneRepoOptions) error {
```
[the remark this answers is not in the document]
mafredri: The option now accepts a depth (true/false/integer) and `initSubmodules` enforces `currentDepth > opts.SubmoduleDepth` at the top of each call. "true" maps to `DefaultSubmoduleDepth` (10).

> 🤖 Posted using `/amend-review` skill via Coder Agents.

testutil/gittest/gittest.go
```diff
@@ -269,6 +269,40 @@ func NewRepo(t *testing.T, fs billy.Filesystem, commits ...CommitFunc) *git.Repo
...
+
+	if opts.TLS {
+		parentSrv = httptest.NewTLSServer(opts.AuthMW(NewServer(fs)))
+	} else {
+		parentSrv = httptest.NewServer(opts.AuthMW(NewServer(fs)))
+	}
+	return parentSrv, submoduleSrv
+}
```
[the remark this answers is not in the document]
mafredri: `CreateGitServerWithSubmodule` is gone; its body is inlined into `TestGitSubmodules`, which now builds the parent and submodule servers directly. `CommitSubmodule` stays exported since it does the gitlink/index work that is genuinely reusable.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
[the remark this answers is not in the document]
mafredri: Settled in-thread: @bjornrobertsson kept the custom resolver to avoid go-git v5's known relative-URL bugs (HTTPS losing a slash via `filepath.Dir`, multi-level `../../` not resolving). Switching to `Submodule.UpdateContext` would reintroduce those.

The SCP-like parent case from @mafredri's later suggestion was adopted via `transport.NewEndpoint` in commit f711af5.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

docs/env-variables.md
```diff
@@ -28,6 +28,7 @@
 | `--git-clone-depth` | `ENVBUILDER_GIT_CLONE_DEPTH` |  | The depth to use when cloning the Git repository. |
 | `--git-clone-single-branch` | `ENVBUILDER_GIT_CLONE_SINGLE_BRANCH` |  | Clone only a single branch of the Git repository. |
 | `--git-clone-thinpack` | `ENVBUILDER_GIT_CLONE_THINPACK` | `true` | Git clone with thin pack compatibility enabled, ensuring that even when thin pack compatibility is activated,it will not be turned on for the domain dev.zaure.com. |
+| `--git-clone-submodules` | `ENVBUILDER_GIT_CLONE_SUBMODULES` |  | Recursively clone Git submodules after cloning the repository. |
```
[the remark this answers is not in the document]
mafredri: Single option with overloaded semantics, as agreed in-thread: `true` -> default depth (10), `false`/`0`/empty -> disabled, positive integer -> max recursion depth.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
mafredri: The suggestion in this thread is empty, and the line it pointed at has changed several times since. The substantive `ResolveSubmoduleURL` discussion was tracked in the AMREM-13 thread (SCP-like URLs) and is now fixed there.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
```diff
@@ -419,3 +433,276 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
...
+		return fmt.Errorf("chroot to .git: %w", err)
+	}
+
+	gitStorage := filesystem.NewStorage(subGitDir, cache.NewObjectLRU(cache.DefaultMaxSize*10))
+
+	// Clone the submodule repository
+	// Use SingleBranch=false to fetch all branches so we can find the commit
+	subRepo, err := git.CloneContext(ctx, gitStorage, subFS, &git.CloneOptions{
+		URL:             resolvedURL,
+		Auth:            opts.RepoAuth,
```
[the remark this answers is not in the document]
mafredri: `SameHost` gating was added in 92f5137. The recursion path that initially still leaked the root credential one level deeper (DEREM-1) is fixed in the follow-up; `initSubmodules` now passes the per-level auth, not the root auth, into recursive calls, and `submoduleAuthFor` is the single decision point.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
mafredri: `RedactURL` is used for every URL we log out of the submodule path: parent URL, `.gitmodules` URL, resolved URL, and the clone-target log line.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

testutil/gittest/gittest.go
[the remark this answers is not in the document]
mafredri: The submodule server is now seeded with `subfile.txt`, and `TestGitSubmodules` reads it back from inside the container (`cat /workspaces/empty/submod/subfile.txt`) so the test fails if anything before the checkout step regresses.

The broader credential-exfiltration scenario you also asked about is covered by the new internal-package tests in `git/submodule_auth_internal_test.go`, which exercise `submoduleAuthFor` and the recursion invariant from DEREM-1. A full multi-host integration test would need a second `httptest` host beyond `127.0.0.1`; happy to add it if you want, otherwise the unit-level test is what we have.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
mafredri: Switched to `transport.NewEndpoint` in commit f711af5, which handles SCP-like URLs and ports. DEREM-7 then closed the residual credential-leak (`Endpoint.String` re-emits embedded userinfo) by stripping `Password` always and `User` for non-`ssh://` protocols before reconstructing the resolved URL.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
mafredri: Fixed. `initSubmodules` now takes the per-level parent URL and parent auth as parameters. Each iteration calls `submoduleAuthFor(parentURL, resolvedURL, parentAuth)`, which is the only place that decides whether to forward; the recursive call passes that decision as the new `parentAuth`. Once `submoduleAuthFor` returns nil at any level, every deeper level sees `parentAuth == nil` and the root credential cannot reappear, regardless of how the deeper hosts match.

This is verified by `TestSubmoduleAuthChainStaysWithheld` in `git/submodule_auth_internal_test.go`.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
mafredri: Verified against native git that `./extras/tool.git` relative to `https://example.com/org/main.git` resolves to `https://example.com/org/main.git/extras/tool.git`, matching the current code:

```
$ # parent remote: https://example.com/org/main.git
$ cat .gitmodules
[submodule "extras"]
        path = extras
        url = ./extras/tool.git

$ git submodule init && git config submodule.extras.url
https://example.com/org/main.git/extras/tool.git
```

go-git uses the same rule (`vendor/.../submodule.go`: `rootEndpoint.Path = path.Join(rootEndpoint.Path, moduleEndpoint.Path)`). Both `.` and `..` are resolved against the parent URL via `path.Join`, not against the parent's directory: `..` works because `path.Join("/org/main.git", "../deps/lib.git")` cleans to `/org/deps/lib.git`. The existing `relativeChild` test expectation is correct.

Closing as not a bug; happy to revisit if you have a reproducer showing a different behavior from upstream git.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
mafredri: Both call sites now use `errors.Is(err, git.NoErrAlreadyUpToDate)`.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git_test.go
mafredri: Removed the two captures introduced in this PR (`git_test.go:674` and `git_test.go:772`). The one at `git_test.go:91` predates this PR and is left untouched.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
mafredri: The "already cloned" path now chroots to `.git` before constructing the storage, matching the fresh-clone path and `CloneRepo` itself.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
mafredri: `CloneRepo` now runs `initSubmodules` whenever `opts.SubmoduleDepth > 0`, not only after a fresh clone. Combined with the DEREM-5 fix (idempotent "already cloned" path), a transient failure during first setup is retried on the next workspace start.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
mafredri: `ResolveSubmoduleURL` now clears credentials before reconstructing the resolved URL. `parentEP.Password` is always cleared; `parentEP.User` is cleared for every protocol except `ssh://`, where the user is the SSH login name rather than a credential. Covered by the new `httpsParentWithTokenStripped`, `httpsParentWithUserPassStripped`, `sshSchemeUserPreserved`, and `sshSchemePasswordStripped` cases.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
mafredri: Nested recursion no longer goes through `sub.Repository()`. A new `openSubmoduleRepo` helper opens the on-disk repo at `parentWorktree/submodulePath/.git` directly, which is the layout `cloneSubmodule` writes. The `sub.Init()` registration requirement is sidestepped.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
mafredri: Submodule `CloneOptions` now sets `Depth: opts.Depth`. If the expected commit is not reachable from the shallow tip, the existing fetch-by-hash fallback (now in `checkoutSubmoduleCommit`) deepens as needed.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git_test.go
mafredri: Partial. The credential-gating logic, which was the highest-risk untested area, is now covered by `TestSubmoduleAuthFor` and `TestSubmoduleAuthChainStaysWithheld` in `git/submodule_auth_internal_test.go`. The retry chain inside `checkoutSubmoduleCommit` (commit-not-found -> direct fetch -> fetch-all -> verify) is still only exercised by the integration test's happy path.

A second `httptest` host on a different name would let us cover the host-mismatch credential-withholding behavior at integration level too. Happy to add that if you want it as part of this PR; otherwise I would track it as a follow-up.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
mafredri: The `!errors.Is(err, git.ErrRepositoryAlreadyExists)` guard is gone. The fresh-clone branch is only entered after the `.git` stat check fails, so `CloneContext` is now expected to succeed and any error is fatal.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
mafredri: The `SameHost` doc comment now states that the port is ignored on purpose to match git's credential-helper convention. No behavior change; this is intentional ecosystem alignment, but it is now explicit in the source.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
mafredri: Added `select { case <-ctx.Done(): return ctx.Err(); default: }` at the top of the submodule loop body in `initSubmodules`.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

options/options.go
mafredri: Renamed the Go field to `Options.GitCloneSubmoduleDepth` so the `int` reads as a recursion depth, aligning with the internal `SubmoduleDepth` type and `CloneRepoOptions.SubmoduleDepth`.

Kept `--git-clone-submodules` and `ENVBUILDER_GIT_CLONE_SUBMODULES` unchanged for user-facing continuity; the flag/env naturally accepts `true`/`false`/integer, while the Go field's name now reflects what the integer represents.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
mafredri: Both checkout paths now produce `"checkout expected commit <hash>: <err>"`. The fresh-clone path already did; the once-cloned path now does too, via the shared `checkoutSubmoduleCommit` helper.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
mafredri: Errors from `openSubmoduleRepo` and from `subWorktree.Submodules()` are now logged at warning level before the loop `continue`s. The old silent `if err == nil` path is gone.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git_test.go
mafredri: Added the `unparseableParent` case (`https://[bad-bracket`, which `transport.NewEndpoint` rejects via `net/url`) so the `expectErr` field is now exercised and the resolver's error branch is covered.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

options/options.go
mafredri: `SubmoduleDepth.Set` now passes `lower` (already trimmed and lowered) into `strconv.Atoi`, so values like `" 5 "` parse. Covered by the new `submodule depth/integer with whitespace` subtest.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

options/options_test.go
mafredri: Submodule-depth cases moved out of the `bool` subtest into a sibling `submodule depth` subtest (`true`, `integer`, `integer with whitespace`, `false`). Test output is now `TestEnvOptionParsing/submodule_depth/...`.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
```diff
@@ -111,28 +116,48 @@ func CloneRepo(ctx context.Context, logf func(string, ...any), opts CloneRepoOpt
...
+			InsecureSkipTLS: opts.Insecure,
+			Depth:           opts.Depth,
+			SingleBranch:    opts.SingleBranch,
+			CABundle:        opts.CABundle,
+			ProxyOptions:    opts.ProxyOptions,
+		})
+		if errors.Is(err, git.ErrRepositoryAlreadyExists) {
+			// The repository was created between our Open and CloneContext
+			// calls. Reopen it so submodule initialization can still run.
+			repo, err = git.Open(fsStorage, gitDir)
```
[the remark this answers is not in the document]
mafredri: Both `git.Open` calls now use `(gitStorage, fs)`, matching the argument order at `CloneContext` and in the submodule code. The unused `fsStorage` variable is removed.

The `AlreadyCloned` test now clones first (producing the `.git/` layout `CloneRepo` expects) then verifies the second call returns `(false, nil)`. Cases where the initial clone is expected to fail skip the subtest.

Fix in #518.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
```diff
@@ -419,3 +445,384 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
...
+
+// submoduleAuthFor returns the auth to use when fetching a submodule. It
+// returns parentAuth if the submodule shares the parent's host, and nil
+// otherwise. A warning is logged when parent auth is set but withheld
+// because the hosts differ.
+func submoduleAuthFor(logf func(string, ...any), parentURL, submoduleURL string, parentAuth transport.AuthMethod) transport.AuthMethod {
+	if parentAuth == nil {
+		return nil
+	}
+	if SameHost(parentURL, submoduleURL) {
```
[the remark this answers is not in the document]
mafredri: Documented as intentional in the `submoduleAuthFor` doc comment. The host-only check forwards parent auth regardless of protocol. When the protocols are incompatible (e.g., SSH auth to an HTTPS submodule), go-git rejects it at the transport layer with `transport.ErrInvalidAuthMethod`. Returning nil here would silently skip auth for a submodule that legitimately needs it under a different protocol.

Fix in #518.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
```diff
@@ -419,3 +445,384 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
...
+			return fmt.Errorf("open existing submodule: %w", err)
+		}
+		return checkoutSubmoduleCommit(ctx, logf, subRepo, expectedHash, submoduleAuth, opts)
+	}
+
+	// Clone the submodule
+	logf("  Cloning submodule from: %s", RedactURL(resolvedURL))
+
+	// Create .git directory for the submodule
+	if err := subFS.MkdirAll(".git", 0o755); err != nil {
```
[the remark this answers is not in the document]
mafredri: If `CloneContext` fails, the `.git` directory is now removed via `billyutil.RemoveAll(subFS, ".git")` before returning the error. The next workspace start will retry with a clean state.

Fix in #518.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
```diff
@@ -419,3 +445,384 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
...
+	// Split the submodule URL into components
+	// and manually walk up the directory tree for each ../
+	currentPath := parentPath
+	relativeParts := strings.Split(submoduleURL, "/")
+
+	for _, part := range relativeParts {
+		if part == ".." {
+			// Go up one directory
+			currentPath = path.Dir(currentPath)
+		} else if part == "." {
```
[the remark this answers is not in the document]
mafredri: Re-raise of DEREM-2 from R1. The claim that `./` should resolve relative to the parent's directory rather than the parent's path contradicts native git's behavior, which was tested in R1:

```
$ git submodule init && git config submodule.extras.url
https://example.com/org/main.git/extras/tool.git
```

go-git resolves the same way (`submodule.go:152`). No change.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
```diff
@@ -111,28 +116,48 @@ func CloneRepo(ctx context.Context, logf func(string, ...any), opts CloneRepoOpt
...
+			alreadyCloned = true
+		}
+		if err != nil {
+			return false, fmt.Errorf("clone %q: %w", opts.RepoURL, err)
+		}
 	}
-	if err != nil {
-		return false, fmt.Errorf("clone %q: %w", opts.RepoURL, err)
+
+	// Initialize submodules on every call, not only after a fresh clone, so
```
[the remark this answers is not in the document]
mafredri: PR description updated via `amend-review update-pr`. The behavior section now reads:
- Submodule initialization runs on every workspace start when enabled, so a transient failure during the first run is automatically retried on the next start
- Already-cloned submodules are detected and only their expected commit is checked out, making re-runs cheap

> 🤖 Posted using `/amend-review` skill via Coder Agents.

integration/integration_test.go
```diff
@@ -418,6 +419,42 @@ func TestSucceedsGitAuth(t *testing.T) {
 	require.Contains(t, gitConfig, srv.URL)
 }
 
+func TestGitSubmodules(t *testing.T) {
```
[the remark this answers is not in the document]
mafredri: Tracked as a follow-up. The nested submodule integration test requires a three-repo setup (parent → submodule → nested submodule) and a Docker container to run. The unit-level `TestSubmoduleAuthChainStaysWithheld` covers the recursion invariant, and the single-level integration test covers the end-to-end happy path.

https://github.com/coder/envbuilder/issues/74

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/submodule_auth_internal_test.go
```diff
@@ -0,0 +1,88 @@
...
+			parentAuth:   parentAuth,
+			wantAuth:     parentAuth,
+		},
+	}
+
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			t.Parallel()
+			var buf bytes.Buffer
+			logf := func(format string, _ ...any) { _, _ = buf.WriteString(format) }
```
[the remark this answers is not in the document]
mafredri: Test `logf` now uses `fmt.Fprintf(&buf, format, args...)` so format args are expanded and the warning check validates the actual output.

Fix in #518.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
```diff
@@ -419,3 +445,384 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
...
+	}
+
+	// For relative URLs, we need to resolve them against the parent's path.
+	// The parent path represents a repository (like a file in filesystem terms),
+	// so ../something means "sibling repository".
+	parentPath := strings.TrimSuffix(parentEP.Path, "/")
+
+	// Split the submodule URL into components
+	// and manually walk up the directory tree for each ../
+	currentPath := parentPath
```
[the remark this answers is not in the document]
mafredri: Same claim as AMREM-17 and DEREM-2 (R1). Already contested with a native git reproduction. No change.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
```diff
@@ -419,3 +445,384 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
...
+			},
+			Auth:            submoduleAuth,
+			Progress:        opts.Progress,
+			InsecureSkipTLS: opts.Insecure,
+			CABundle:        opts.CABundle,
+			ProxyOptions:    opts.ProxyOptions,
+		})
+		if fetchErr != nil && !errors.Is(fetchErr, git.NoErrAlreadyUpToDate) {
+			// If that fails, try fetching all refs
+			logf("  Direct fetch failed, fetching all refs...")
```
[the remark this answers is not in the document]
mafredri: Log line now includes the error: `logf("  Direct fetch failed (%v), fetching all refs...", fetchErr)`.

Fix in #518.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
```diff
@@ -419,3 +445,384 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
...
+		if len(nestedSubs) == 0 {
+			continue
+		}
+		logf("  Found %d nested submodule(s) in %s", len(nestedSubs), subConfig.Name)
+		if err := initSubmodules(ctx, logf, subRepo, subWorktree, resolvedURL, submoduleAuth, opts, currentDepth+1); err != nil {
+			return fmt.Errorf("init nested submodules in %q: %w", subConfig.Name, err)
+		}
+	}
+
+	logf("✓ All submodules initialized successfully")
```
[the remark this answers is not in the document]
mafredri: `initSubmodules` now tracks a warning counter. The final log line says `"⚠ Submodule initialization finished with N warning(s)"` when warnings were emitted, and `"✓ All submodules initialized successfully"` only when none were.

Fix in #518.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
```diff
@@ -419,3 +445,384 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
...
 	}
 }
+
+// scpLikeURLRegex matches SCP-like URLs: user@host:path (where host is not empty and path doesn't start with /)
+// This handles: git@github.com:org/repo, deploy@host:repo, user@10.0.0.5:project
+var scpLikeURLRegex = regexp.MustCompile(`^([^@]+)@([^:]+):(.+)$`)
+
+// extractHost extracts the host from a URL, handling both standard URLs and SCP-like URLs.
+// Returns empty string if host cannot be determined.
+func extractHost(u string) string {
```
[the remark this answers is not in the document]
mafredri: `extractHost` now delegates to `transport.NewEndpoint` instead of reimplementing SCP URL parsing. This ensures the same parser is used by both `SameHost` and `ResolveSubmoduleURL`.

Fix in #518.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
```diff
@@ -419,3 +445,384 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
...
+// submoduleAuth may be nil even when the parent was authenticated.
+func cloneSubmodule(ctx context.Context, logf func(string, ...any), parentWorktree *git.Worktree, subConfig *config.Submodule, expectedHash plumbing.Hash, resolvedURL string, submoduleAuth transport.AuthMethod, opts CloneRepoOptions) error {
+	// Get the submodule directory within the parent worktree
+	subFS, err := parentWorktree.Filesystem.Chroot(subConfig.Path)
+	if err != nil {
+		return fmt.Errorf("chroot to submodule path: %w", err)
+	}
+
+	// Check if already cloned
+	if _, statErr := subFS.Stat(".git"); statErr == nil {
```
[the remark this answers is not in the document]
mafredri: The already-cloned branch in `cloneSubmodule` now calls `openSubmoduleRepo(parentWorktree, subConfig.Path)` instead of duplicating the Chroot/NewStorage/Open sequence.

Fix in #518.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
```diff
@@ -419,3 +445,384 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
...
+			Auth:            submoduleAuth,
+			Progress:        opts.Progress,
+			InsecureSkipTLS: opts.Insecure,
+			CABundle:        opts.CABundle,
+			ProxyOptions:    opts.ProxyOptions,
+		})
+		if fetchErr != nil && !errors.Is(fetchErr, git.NoErrAlreadyUpToDate) {
+			// If that fails, try fetching all refs
+			logf("  Direct fetch failed, fetching all refs...")
+			fetchAllErr := subRepo.FetchContext(ctx, &git.FetchOptions{
```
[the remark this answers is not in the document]
mafredri: The depth-propagation log line is a good idea but is a follow-up. The current behavior (shallow clone + fetch-by-hash fallback) is correct; the missing log just makes it harder for operators to diagnose slow starts.

https://github.com/coder/envbuilder/issues/74

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
```diff
@@ -419,3 +445,384 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
...
+	if matches := scpLikeURLRegex.FindStringSubmatch(u); matches != nil {
+		return strings.ToLower(matches[2])
+	}
+
+	return ""
+}
+
+// SameHost checks if two URLs point to the same host.
+// Used to determine if credentials should be forwarded to submodules.
+// The comparison is hostname-only. Port is ignored to match git's
```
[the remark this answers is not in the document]
mafredri: Doc comment corrected. Port is ignored as a simplification, not to match git credential helpers (which include port in their key).

Fix in #518.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

options/options.go
```diff
@@ -5,13 +5,53 @@ import (
...
+	case "true", "yes":
+		*s = DefaultSubmoduleDepth
+		return nil
+	case "false", "no", "":
+		*s = 0
+		return nil
+	}
+	n, err := strconv.Atoi(lower)
+	if err != nil {
+		return fmt.Errorf("invalid submodule depth %q: must be true, false, or a positive integer", val)
```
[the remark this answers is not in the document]
mafredri: All four occurrences now say "non-negative integer" and the CLI description clarifies "0 disables, max 100".

Fix in #518.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
```diff
@@ -419,3 +445,384 @@ func ProgressWriter(write func(line string, args ...any)) io.WriteCloser {
...
+		return fmt.Errorf("clone submodule repository: %w", err)
+	}
+
+	return checkoutSubmoduleCommit(ctx, logf, subRepo, expectedHash, submoduleAuth, opts)
+}
+
+// checkoutSubmoduleCommit ensures expectedHash is present in subRepo,
+// fetching it from the remote if it is not already there, and then checks
+// it out into the submodule's worktree.
+func checkoutSubmoduleCommit(ctx context.Context, logf func(string, ...any), subRepo *git.Repository, expectedHash plumbing.Hash, submoduleAuth transport.AuthMethod, opts CloneRepoOptions) error {
```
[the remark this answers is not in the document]
mafredri: The three code paths (commit present, fetch-by-hash, fetch-all) in `checkoutSubmoduleCommit` are only exercised by the integration happy path. Adding unit tests for the fetch fallback paths requires a git server that can simulate a shallow clone where the pinned commit is not the tip. This is a follow-up.

https://github.com/coder/envbuilder/issues/74

> 🤖 Posted using `/amend-review` skill via Coder Agents.

integration/integration_test.go
```diff
@@ -418,6 +419,42 @@ func TestSucceedsGitAuth(t *testing.T) {
...
+	t.Parallel()
+
+	submoduleFS := memfs.New()
+	submoduleRepo := gittest.NewRepo(t, submoduleFS,
+		gittest.Commit(t, "subfile.txt", "submodule content", "submodule commit"),
+	)
+	submoduleHead, err := submoduleRepo.Head()
+	require.NoError(t, err)
+	submoduleSrv := httptest.NewServer(mwtest.BasicAuthMW("", "")(gittest.NewServer(submoduleFS)))
```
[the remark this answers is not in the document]
mafredri: An authenticated integration test (where the server rejects unauthenticated requests) would cover the full auth forwarding chain. The unit tests cover the decision logic; the gap is the HTTP layer. Follow-up.

https://github.com/coder/envbuilder/issues/74

> 🤖 Posted using `/amend-review` skill via Coder Agents.

options/options.go
```diff
@@ -5,13 +5,53 @@ import (
...
+	}
+	if n < 0 {
```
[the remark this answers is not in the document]
mafredri: `SubmoduleDepth.Set` now rejects values above `MaxSubmoduleDepth` (100) with a clear error.

Fix in #518.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

tomelliff:
Thanks again for this @bjornrobertsson. Any chance this can get a release so we can drop our internally hosted build from this branch at some point please?

PR #507, opened 2026-04-10
fix(devcontainer): use 0o755 for feature directories (missing execute bit)

### LINKED ISSUE #506: bug: MkdirAll uses 0o644 for feature directories (missing execute bit)
`devcontainer/devcontainer.go` calls `fs.MkdirAll(featuresDir, 0o644)` and `fs.MkdirAll(featureDir, 0o644)` when creating feature extraction directories. `0o644` is a file permission mask — directories need the execute bit (`0o755`) to be traversable by non-root users.

On the in-memory filesystem used in tests this is harmless, but on a real filesystem it could prevent non-root container users from accessing feature install scripts.

This pre-dates #500 (the feature ordering PR) — the same `0o644` was used in the original code.

### Fix

Change both `MkdirAll` calls to use `0o755`.

> 🤖 Written by a Coder Agent.

johnstcn:
- Change `MkdirAll` calls for `featuresDir` and per-feature `featureDir` from `0o644` to `0o755`.
- Directories need the execute bit to be traversable by non-root users.

Fixes #506

> 🤖 Written by a Coder Agent. Will be reviewed by a human.

### DIFF
--- a/devcontainer/devcontainer.go
+++ b/devcontainer/devcontainer.go
@@ -227,7 +227,7 @@ func (s *Spec) compileFeatures(fs billy.Filesystem, devcontainerDir, scratchDir 
 	}
 
 	featuresDir := filepath.Join(scratchDir, "features")
-	err := fs.MkdirAll(featuresDir, 0o644)
+	err := fs.MkdirAll(featuresDir, 0o755)
 	if err != nil {
 		return "", nil, fmt.Errorf("create features directory: %w", err)
 	}
@@ -277,7 +277,7 @@ func (s *Spec) compileFeatures(fs billy.Filesystem, devcontainerDir, scratchDir 
 		featureSha := md5.Sum([]byte(featureRefRaw))
 		featureName := filepath.Base(featureRef)
 		featureDir := filepath.Join(featuresDir, fmt.Sprintf("%s-%x", featureName, featureSha[:4]))
-		if err := fs.MkdirAll(featureDir, 0o644); err != nil {
+		if err := fs.MkdirAll(featureDir, 0o755); err != nil {
 			return "", nil, err
 		}
 		spec, err := features.Extract(fs, devcontainerDir, featureDir, featureRefRaw)
--- a/devcontainer/devcontainer.go
+++ b/devcontainer/devcontainer.go
@@ -278,7 +278,7 @@ func (s *Spec) compileFeatures(fs billy.Filesystem, devcontainerDir, scratchDir 
 		featureName := filepath.Base(featureRef)
 		featureDir := filepath.Join(featuresDir, fmt.Sprintf("%s-%x", featureName, featureSha[:4]))
 		if err := fs.MkdirAll(featureDir, 0o755); err != nil {
-			return "", nil, err
+			return "", nil, fmt.Errorf("create feature directory for %s at %s: %w", featureRefRaw, featureDir, err)
 		}
 		spec, err := features.Extract(fs, devcontainerDir, featureDir, featureRefRaw)
 		if err != nil {
--- a/devcontainer/devcontainer_test.go
+++ b/devcontainer/devcontainer_test.go
@@ -115,6 +115,15 @@ WORKDIR `+featureTwoDir+`
 ENV POTATO=example
 RUN VERSION="potato" _CONTAINER_USER="1000" _REMOTE_USER="1000" ./install.sh
 USER 1000`, params.DockerfileContent)
+
+		// Verify feature directories are created with 0o755 (not
+		// 0o644) so non-root container users can traverse them.
+		for _, dir := range []string{workingDir + "/features", featureOneDir, featureTwoDir} {
+			info, err := fs.Stat(dir)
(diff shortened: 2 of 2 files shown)


devcontainer/devcontainer.go
```diff
@@ -227,7 +227,7 @@ func (s *Spec) compileFeatures(fs billy.Filesystem, devcontainerDir, scratchDir
 	}
 
 	featuresDir := filepath.Join(scratchDir, "features")
-	err := fs.MkdirAll(featuresDir, 0o644)
+	err := fs.MkdirAll(featuresDir, 0o755)
```
mafredri, line 230: **P2** No test verifies the permission fix that is the entire point of this PR. *(Bisky P2, Hisoka P3, Mafu-san P3, Mafuuu P3)*

> "`TestCompileWithFeatures` uses `memfs.New()`. memfs stores file modes faithfully, but the test never calls `Stat` on the created directories, never checks `Mode()`, never asserts anything about permissions. The test passes identically with `0o644` or `0o755`." -- Bisky

memfs stores the mode argument and returns it via `Stat().Mode()`. The integration test runs features as `USER root` (line 299 of this file), so root bypasses permission checks entirely. Bug #506 was user-found, not test-found. The same discovery path will repeat if this regresses.

A minimal regression test (using the existing memfs infrastructure):

```go
fi, err := fs.Stat(featuresDir)
require.NoError(t, err)
require.Equal(t, os.FileMode(0o755), fi.Mode().Perm(),
    "feature directories must be traversable by non-root users")
```

> 🤖

devcontainer/devcontainer.go
```diff
@@ -277,7 +277,7 @@ func (s *Spec) compileFeatures(fs billy.Filesystem, devcontainerDir, scratchDir
 		featureSha := md5.Sum([]byte(featureRefRaw))
 		featureName := filepath.Base(featureRef)
 		featureDir := filepath.Join(featuresDir, fmt.Sprintf("%s-%x", featureName, featureSha[:4]))
-		if err := fs.MkdirAll(featureDir, 0o644); err != nil {
+		if err := fs.MkdirAll(featureDir, 0o755); err != nil {
```
mafredri, line 280: **P4** Pre-existing: error from this `MkdirAll` is returned unwrapped, unlike the parent `featuresDir` creation at line 232 which wraps with `fmt.Errorf("create features directory: %w", err)`. When this fails, the caller gets a bare OS error with no indication of which feature or operation failed. *(Mafuuu)*

Not introduced by this PR. Noting because it's adjacent to the changed line.

> 🤖

devcontainer/devcontainer.go
```diff
@@ -277,7 +277,7 @@ func (s *Spec) compileFeatures(fs billy.Filesystem, devcontainerDir, scratchDir
 		featureSha := md5.Sum([]byte(featureRefRaw))
```
mafredri, line 277: **P4** Pre-existing: `featureSha` uses `md5.Sum`, not a SHA-family hash. The name misleads about collision and length properties. `featureHash` would be accurate. *(Gon)*

Not introduced by this PR.

> 🤖

mafredri (commented):
Correct fix, well-scoped. Two constants changed, two-sentence description, no scope creep. The permission values now match every other production `MkdirAll` in the codebase (all use `0o755`, `0o750`, or `0o700`).

One concern (P2), two adjacent nits (P4). The P2 is that the fix ships without proof it works: no test asserts the permission value, memfs doesn't enforce permissions, and the integration test runs as root. Five of seven reviewers flagged this independently.

> "Revert the fix, all tests stay green. The bug described in #506 has no regression test." -- Bisky

The two P4s are pre-existing issues adjacent to the changed lines, not introduced by this PR.

devcontainer/devcontainer.go
johnstcn: Also raised by copilot, addressed below.

devcontainer/devcontainer.go
johnstcn: Addressed in 3304543adbddcc961d235bacf413934071268be1

mafredri (commented):
Round 2. F1 (P2, missing regression test) addressed in 3304543 with permission assertions. F2 (P4, unwrapped error) addressed in 670d34b with descriptive wrapping. Both look good.

F3 (P4, `featureSha` naming) is unresolved: thread was closed with no reply and no code change. This is a pre-existing nit, not introduced by this PR. A brief acknowledgment (e.g. "won't fix, pre-existing") would unblock the next review round. Holding further review until then.

devcontainer/devcontainer.go
mafredri: This thread was resolved without a reply or code change. A quick acknowledgment would unblock the next review round. Since this is pre-existing and P4, "won't fix" is a perfectly fine response.

> 🤖

johnstcn: acknowledging pre-existing issue. 

PS Note for maintainer: P4 I would qualify as a non-blocking "nit".

PR #515, opened 2026-05-20
test: cover cross-host auth forwarding for submodule clones

mafredri:
Adds a test that verifies the SameHost auth gating in `cloneSubmodule` actually behaves as intended end-to-end. The submodule HTTP server is wrapped with an observer that records every request carrying an `Authorization` header; the same-host subtest expects auth to be forwarded, the cross-host subtest expects it to be withheld.

The cross-host case uses `localhost` for the URL in `.gitmodules` and `127.0.0.1` for the parent server's URL. Both resolve to the loopback interface so the submodule clone still succeeds, but `SameHost` compares hostname strings and reports a mismatch, so the auth must not be sent.

Covers the security half of mafredri's thread PRRT_kwDOJPSNMM5vAH0h, where mafredri asked whether the gittest helper should be expanded to exercise the auth-exfiltration guard.

<details>
<summary>Why this PR and not the changes I considered</summary>

I also investigated AMREM-7 (move `cloneSubmodule` over to go-git's native `Submodule.Update()`). go-git v5.16.5's `Submodule.Init` resolves relative URLs via `transport.NewEndpoint`, which absolutizes against CWD, so the v5 implementation is broken for `https://parent + ../sibling` style submodules. The v6 fix in `go-git/go-git@0e6f8484` lives in `Worktree.Submodules()` and is not backported. We could pre-resolve URLs ourselves and call native `Update`, but in v5 `SubmoduleUpdateOptions` does not surface `InsecureSkipTLS`, `CABundle`, or `ProxyOptions`, and the internal `fetchAndCheckout` only forwards `Auth` and `Depth` to `FetchContext`. Going native silently drops self-signed-cert, corporate-CA, and proxy support for submodule fetches. Keeping the bespoke clone is the right call; the reasoning is the TLS/proxy regression, not the relative-URL bug bjornrobertsson cited in-thread.

</details>

> 🤖 This PR was created with the help of Coder Agents, and _will be_ reviewed by a human. 🏂🏻

### DIFF
--- a/git/git_test.go
+++ b/git/git_test.go
@@ -1142,4 +1142,3 @@ func TestCloneRepoSubmoduleHostAuth(t *testing.T) {
 		})
 	}
 }
-
--- a/git/git_test.go
+++ b/git/git_test.go
@@ -5,11 +5,13 @@ import (
 	"crypto/ed25519"
 	"encoding/base64"
 	"io"
+	"net/http"
 	"net/http/httptest"
 	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"testing"
 
 	"github.com/coder/envbuilder/git"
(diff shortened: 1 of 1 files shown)


johnstcn:
needs rebase

johnstcn (approved):

PR #518, opened 2026-05-22
fix(git): fix submodule init on workspace restart

mafredri:
Patch PR for #485 (round 2) that addresses the second review pass. Targets the PR's branch.

## Changes

**P1: `git.Open` arguments swapped (AMREM-14)**

Both calls at lines 112 and 136 passed `(fsStorage, gitDir)` when the correct order is `(storer-backed-by-.git, worktree-fs)`. This made every workspace restart fail to detect the existing repo, enter the clone path, and hit `ErrRepositoryAlreadyExists`. The `AlreadyCloned` test is updated to actually clone first (producing the right on-disk layout) instead of using `gittest.NewRepo` which writes objects at the root.

**P2: Orphaned `.git` blocks retries (AMREM-16)**

If `CloneContext` fails (network timeout, auth error), the `.git` directory created by `MkdirAll` persisted. On the next start, `Stat(".git")` took the already-cloned fast path and tried to `Open` a corrupt repo. Now the `.git` directory is removed on clone failure.

**P3 fixes**

- `extractHost` rewritten to use `transport.NewEndpoint`, removing the duplicate SCP regex parser (AMREM-24).
- Already-cloned branch in `cloneSubmodule` now calls `openSubmoduleRepo` instead of duplicating the Chroot/NewStorage/Open sequence (AMREM-25).
- `fetchErr` included in the "Direct fetch failed" log message (AMREM-22).
- `initSubmodules` tracks warnings and adjusts the final log line accordingly (AMREM-23).
- `SameHost` doc corrected: port is ignored as a simplification, not to match git credential helpers (AMREM-27).
- "positive integer" changed to "non-negative integer (0 disables)" everywhere (AMREM-28).
- `SubmoduleDepth` capped at 100 (AMREM-31).
- Cross-protocol auth forwarding documented as intentional in `submoduleAuthFor` (AMREM-15).
- Test `logf` fixed to use `fmt.Fprintf` so format args are expanded (AMREM-20).

**Docs: PR description updated (AMREM-18)**

The behavior section now accurately describes the retry-on-every-start semantics.

## Contested

**AMREM-17 and AMREM-21** re-raise the `./`-relative URL resolution claim from DEREM-2 (R1), which was already contested with a native git reproduction showing the current behavior is correct. No change.

> 🤖 This PR was created with the help of Coder Agents, and _will be_ reviewed by a human. 🏂🏻

### DIFF
--- a/docs/env-variables.md
+++ b/docs/env-variables.md
@@ -28,7 +28,7 @@
 | `--git-clone-depth` | `ENVBUILDER_GIT_CLONE_DEPTH` |  | The depth to use when cloning the Git repository. |
 | `--git-clone-single-branch` | `ENVBUILDER_GIT_CLONE_SINGLE_BRANCH` |  | Clone only a single branch of the Git repository. |
 | `--git-clone-thinpack` | `ENVBUILDER_GIT_CLONE_THINPACK` | `true` | Git clone with thin pack compatibility enabled, ensuring that even when thin pack compatibility is activated,it will not be turned on for the domain dev.azure.com. |
-| `--git-clone-submodules` | `ENVBUILDER_GIT_CLONE_SUBMODULES` |  | Clone Git submodules after cloning the repository. Accepts 'true' (max depth 10), 'false' (disabled), or a positive integer for max recursion depth. |
+| `--git-clone-submodules` | `ENVBUILDER_GIT_CLONE_SUBMODULES` |  | Clone Git submodules after cloning the repository. Accepts 'true' (max depth 10), 'false' (disabled), or a non-negative integer for max recursion depth (0 disables, max 100). |
 | `--git-username` | `ENVBUILDER_GIT_USERNAME` |  | The username to use for Git authentication. This is optional. |
 | `--git-password` | `ENVBUILDER_GIT_PASSWORD` |  | The password to use for Git authentication. This is optional. |
 | `--git-ssh-private-key-path` | `ENVBUILDER_GIT_SSH_PRIVATE_KEY_PATH` |  | Path to an SSH private key to be used for Git authentication. If this is set, then GIT_SSH_PRIVATE_KEY_BASE64 cannot be set. |
--- a/git/git.go
+++ b/git/git.go
@@ -17,6 +17,7 @@ import (
 	"github.com/coder/envbuilder/options"
 
 	"github.com/go-git/go-billy/v5"
+	billyutil "github.com/go-git/go-billy/v5/util"
 	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/config"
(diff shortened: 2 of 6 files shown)


docs/env-variables.md
```diff
@@ -28,7 +28,7 @@
 | `--git-clone-depth` | `ENVBUILDER_GIT_CLONE_DEPTH` |  | The depth to use when cloning the Git repository. |
 | `--git-clone-single-branch` | `ENVBUILDER_GIT_CLONE_SINGLE_BRANCH` |  | Clone only a single branch of the Git repository. |
 | `--git-clone-thinpack` | `ENVBUILDER_GIT_CLONE_THINPACK` | `true` | Git clone with thin pack compatibility enabled, ensuring that even when thin pack compatibility is activated,it will not be turned on for the domain dev.azure.com. |
-| `--git-clone-submodules` | `ENVBUILDER_GIT_CLONE_SUBMODULES` |  | Clone Git submodules after cloning the repository. Accepts 'true' (max depth 10), 'false' (disabled), or a positive integer for max recursion depth. |
+| `--git-clone-submodules` | `ENVBUILDER_GIT_CLONE_SUBMODULES` |  | Clone Git submodules after cloning the repository. Accepts 'true' (max depth 10), 'false' (disabled), or a non-negative integer for max recursion depth (0 disables, max 100). |
```
johnstcn, line 31: 100? That's a _lot_ of submodules XD

mafredri: I'm guessing even that isn't going to be enough for that one user 🙈

git/git.go
```diff
@@ -768,6 +766,7 @@ func cloneSubmodule(ctx context.Context, logf func(string, ...any), parentWorktr
 		NoCheckout:      true,
 	})
 	if err != nil {
+		_ = billyutil.RemoveAll(subFS, ".git")
```
johnstcn: nit: if we fail to remove we could leave things silently in a broken state. Suggest at least logging the error as a warning here.

johnstcn (approved):

docs/env-variables.md
mafredri: @mafredri already replied in-thread. The cap is a safeguard against runaway recursion, not an expected operating depth. Happy to lower it if the team prefers a tighter default.

> 🤖 Posted using `/amend-review` skill via Coder Agents.

git/git.go
mafredri: Now logs a warning when the cleanup fails:
```go
if rmErr := billyutil.RemoveAll(subFS, ".git"); rmErr != nil {
    logf("  ⚠ Failed to clean up .git after clone failure: %v", rmErr)
}
```

> 🤖 Posted using `/amend-review` skill via Coder Agents.

